Issue #15, opened 2017-04-06
Broken on VSCode 1.11.0


marekr:
Attempting to open a preview window on 1.11.0 causes a error.

"Cannot read property 'document' of undefined."

This is probably due to the new workspaces feature. Really sucks how VSCode can't maintain API compatibility worth a damn due to JS land.

joaompinto:
Hello,
it's working for me on 1.11, are you using Linux or Windows ?

marekr:
Windows.

I just updated on a second PC witht he same result.

fabianlaule:
Same here with 1.11.2 and OS X

sanguine8082:
works ok for me.  

Windows 7 1.11.2

brettenbacher:
Same happens with my installations:
Tried out 1.11.2 and 1.12.0 (insider) on Ubuntu 16.04 and 1.11.2 on OS X (all fresh installs) using Asciidoctor 1.5.5.

FrVaBe:
Preview also fails for me on(Windows 7 with message "Cannot read property 'document' of undefined." 

- VSCode Version: Code 1.12.1 (f6868fce3eeb16663840eb82123369dec6077a9b, 2017-05-04T21:26:50.689Z)
- OS Version: Windows_NT ia32 6.1.7601
- Extensions: 

|Extension|Author|Version|
|---|---|---|
|markdown-toc|AlanWalk|1.5.1|
|asciidoctor-vscode|joaompinto|0.3.3|;

elfgoh:
I confirm this issue on Mac OS, version ` 1.12.2`

joaompinto:
Fixed on 0.3.5 .

sanguine8082:
Scratch that, Open Preview to side works, but Open Preview does not

marekr:
@sanguine8082 have you upgraded to 0.3.5 (you have to manually click upgrade)? It should work in all cases now after I fixed it.

sanguine8082:
Oh, my mistake!  

Definitely works now.  Thanks a bunch!

PR #21, opened 2017-07-01
Make plugin work again on latest vscode

### LINKED ISSUE #20: Wrong preview after opening new .adoc file
When I open a folder and work on one .adoc file and open the live preview everything works fine.
When I switch to another .adoc file in the same project (not even necessarily in the same sub-dir) the preview window remains showing the original .adoc preview.
The title of the preview window matches the open .adoc file that I wanted a preview of, but it is the wrong document.
Occassionally CMD+W on the windows will allow it to correct, but usually I have to exit/restart VSCode.
On Mac 10.12.5
VSCode 1.13.1
AsciiDoc Live Preview extension v0.3.3

marekr:
Fixes #20, #15 

- Updated to latest vscode plugin config item names
- Prevent undefined error when opening window
- Don't register the workspace listener every time we create a textdocumentprovider
- Add preview to side title menu extension
- Fix the open to left/right functionality to always track a document rather than jump around, ditching activeTextEditor fixes the undefined error as well
- vsce requires https images in the readme now to create a vsix

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -2,6 +2,7 @@
 {
 	"version": "0.1.0",
 	"configurations": [
+		
 		{
 			"name": "Launch Extension",
 			"type": "extensionHost",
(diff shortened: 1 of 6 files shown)


joaompinto:
Changes release to the marketplace on 0.3.4, thanks very much.

PR #45, opened 2018-02-16
fixes #27 Please don't use Ctrl-K for previewing

### LINKED ISSUE #27: Please don't use Ctrl-K for previewing
`Ctrl-K` is by default bound to deleting all right on VSCode on Mac (and maybe other platforms). This extension is great but please don't hide the default `Ctrl-K` with the previewing shortcut.

gdamore:
This changes the keybindings to only be active when editing
asciidoc files, and changse them to use appropriate mappings
for mac, as well as avoiding keybindings that are used elsewhere.

I've verified that these keybindings do not conflict with those
used on my system (a mac) although I've not tested against every
possible extension, and I have not tested Windows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,8 +9,8 @@ An extension to preview AsciiDoc text using the _AsciiDoctor_ publishing tool.
 
 The extension can be activated in two ways
 
-* Toggle Preview - `ctrl+shift+r`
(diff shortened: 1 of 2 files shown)


joaompinto:
Just for info, the next release will change the binding to CTRL+K + V to make the  key bindings aligned with the markdown preview extension.

PR #107, opened 2018-09-08
Add option to use default stylesheet

### LINKED ISSUE #98: Option to use default stylesheet
This extension is just getting better all the time. Thanks for your hard work on it.

For my use case I would prefer the output styling to match the default Asciidoctor stylesheet (the way this extension used to work some time ago). This is because I prefer these styles (aesthetically) and also because they will reflect the output which is produced by my build system which uses the default stylesheets.

It would be great if we could have this as an option.

spegoraro:
Add a new setting `useDefaultStylesheet` which, if enabled, renders using the default Asciidoctor stylesheet instead of the extension's default one.

This is compatible with the focus-following and highlighting behaviour.

Resolves #98


joaompinto:
Thanks for the contribution.

Issue #95, opened 2018-07-06
Show heading tree in outline view


bitfield:
Now that outline support in Code is out of preview, it would be great if the Outline view on an Asciidoc file showed the headings in a (foldable) tree, like the Markdown extension:

![screen shot 2018-07-06 at 12 44 19](https://user-images.githubusercontent.com/142262/42377213-4f924332-811a-11e8-9503-417c67f83899.png)

I would be most grateful for this, as I am using Asciidoc to write a book, currently at 12 chapters and 100,000 words, and the outline view is ideal for navigating inside chapters.

naltatis:
I'm not sure which version introduced it and I can't find it in the CHANGELOG. But VSCode now shows proper outlines for .adoc documents 💚💛

![outline](https://user-images.githubusercontent.com/152287/69481714-db9eb400-0e13-11ea-80a6-04f823cc743c.png)

joaompinto:
It was probably fixed in 2.7.7, related to https://github.com/asciidoctor/asciidoctor-vscode/pull/234 .
Thanks for the feedback :)

Issue #87, opened 2018-05-22
Add content navigation view


joaompinto:
Inspired by https://github.com/AlanWalk/markdown-navigation

sagebind:
This is largely the same issue as #95, except utilizing a now built-in feature of Code.

danyill:
I think this is now adequately covered by the symbol/outline view so I'll close this issue.

Feel free to re-open if I have misunderstood.

PR #111, opened 2018-09-22
Add support for tree-like outline

sagebind:
Use the new `DocumentSymbol` API to provide a hierarchical symbol tree in order to have a more useful outline for navigation. This is a huge improvement for working with large AsciiDoc documents.

The parsing routine added here is pretty rudimentary, but it seems to get the job done in my testing and looks quite nice.

This also fixes a small issue where lines that _look_ like headers inside literal blocks showed up in the outline.

Resolves issue #95 and #87.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "icon": "images/icon.png",
     "main": "./out/src/extension",
     "engines": {
-        "vscode": "^1.21.0"
+        "vscode": "^1.25.0"
     },
     "categories": [
         "Programming Languages"
--- a/src/AsciiDocSymbolProvider.ts
+++ b/src/AsciiDocSymbolProvider.ts
@@ -10,7 +10,7 @@ import {
 export default function registerDocumentSymbolProvider(): Disposable {
 
         const _atxPattern = /^((=|#){1,6})\s+(.+)$/;
-        const _blockPattern = /^[.\-]{4}$/;
+        const _blockPattern = /^(\.{4}|-{4})$/;
 
         return languages.registerDocumentSymbolProvider('asciidoc', {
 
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -20,7 +20,7 @@
     "icon": "images/icon.png",
     "main": "./out/src/extension",
     "engines": {
-        "vscode": "^1.21.0"
+        "vscode": "^1.25.0"
```
sagebind: This bump is required to get the new `DocumentSymbol` API.

src/AsciiDocSymbolProvider.ts
```diff
@@ -1,35 +1,77 @@
...
+                        continue;
+                    }
+
+                    const match = _atxPattern.exec(text);
+                    if (match !== null) {
+                        const depth = match[1].length;
+                        const symbol = new DocumentSymbol(
+                            match[3],
+                            null,
+                            SymbolKind.String,
```
sagebind: The `SymbolKind` also changed to match the kind exposed by the built-in Markdown symbol provider.

src/AsciiDocSymbolProvider.ts
```diff
@@ -1,35 +1,77 @@
...
+                    // Ignore lines in a literal block.
+                    if (currentBlock !== null) {
+                        continue;
+                    }
+
+                    const match = _atxPattern.exec(text);
+                    if (match !== null) {
+                        const depth = match[1].length;
+                        const symbol = new DocumentSymbol(
+                            match[3],
```
sagebind: I opted to _not_ include the leading syntax of the header in the outline view. This deviates from the behavior in Markdown, but it seems unnecessarily noisy to me.

So instead of showing `== My Section` in the outline we just show `My Section`.

joaompinto:
@sagebind sorry for the long time I took to review PR, this fix is not applicable to the current version, right now the symbol provider is disabled, the plan is to adapt the code that is available from the markdown plugin. Thanks for your patience and comprehension.

PR #113, opened 2018-09-27
Added compatibility with macOS

fcrespo82:
Added wkhtmlpdf binary for compatibility with macOS.

This should make easier to test my first option from https://github.com/joaompinto/asciidoctor-vscode/issues/79#issuecomment-424566452

If you find its a good idea, I can then extract the binary from the 32 bit version and put there too.


joaompinto:
@fcrespo82 thanks for the contribution. I did a manual merge.

PR #115, opened 2018-10-02
Improved optional paragraph to include informations about PlantUML support

Riduidel:
As there are already two issues about that subject, I thought it could be a good idea to describe how PlantUML support is not such a problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,36 @@ Launch VS Code Quick Open (Ctrl+P), paste the following command, and press enter
 
 ## Optional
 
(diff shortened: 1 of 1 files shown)


mojavelinux:
Actually, PlantUML is supported by Asciidoctor.js if you use https://github.com/eshepelyuk/asciidoctor-plantuml.js.

It works by delegating to a diagram server. There are plans to add something similar to Asciidoctor Diagram so support is not limited to PlantUML, but that hasn't happened yet.

joaompinto:
@Riduidel thanks for the contribution and sorry for the long time to merge it.

@mojavelinux I will look into the asciidoctor.js integration, thanks for the tip.

PR #134, opened 2018-11-30
YAML syntax file, improved inline syntax detection, automated building

MatteoCampinoti94:
As per the pull request title, these commits are aimed at making the syntax detection more precise by using better pattern matching and a YAML file for easier editing. Script files have also been added to automate building.

## syntaxes/Asciidoctor.YAML-tmLanguage
This YAML file contains all the syntax patterns and keywords in a much more friendly format than plist.
Furthermore, instead of using a simple `match` it uses `beginCaptures`, `endCaptures` and `patterns: -include`, thus allowing the extension to match nested syntax (i.e. `*_test_*` is recognized and displayed as a bold italic block).

## script/yaml-to-plist
This Python script simply converts the new YAML file to the .tmLanguage file needed by the extension.

## script/build
This shell scrip automates extensions building:
1. convert `syntaxes/Asciidoctor.YAML-tmLanguage` to `syntaxes/Asciidoctor.tmLanguage`
2. run `npm install`
3. install global dependencies `vsce` and `typescript` if not present
4. package extension using `vsce`
5. if `install` is passed as argument then the extension will be installed (using `code --install-extension *.vsix`)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ node_modules/
 .asciidoctor/
 test/samples/*.pdf
 media/*.js
+package-lock.json
--- a/script/build
+++ b/script/build
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+# Exit if any operation fails
+set -e
+
+# Move to project root
+cd "$(dirname "$0")/.."
+
+# Generate plist syntax
+echo "==> Converting YAML syntax to plist"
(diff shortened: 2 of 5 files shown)


joaompinto:
@MatteoCampinoti94 thanks a lot for the contribution

MatteoCampinoti94:
You're welcome :)

MatteoCampinoti94:
The regex patterns should work correctly, I've put lots of checks in place, but there might be weird situations where they dont work

PR #136, opened 2018-12-03
Typographic quotes, build script overhaul, email pattern fix

MatteoCampinoti94:
# Syntax
Typographic quotes have been added, both ```"`double`"``` and ```'`single`'```.

# script/build
The script now accepts arguments and can do single operations without having to do the whole building:
- `-h`, `--help`, `help`
    1. show help message
- `build` 
    1. install local node dependencies
    2. install global node dependencies
    3. generate the .tmLanguage file
    4. package the extension
- `npm`
    1. install local node dependencies
    2. install global node dependencies
- `npm_local`, `npm_local_deps`
    1. install local node dependencies
- `npm_global`, `npm_global_deps`
    1. install global node dependencie
- `generate`, `generate_syntax`
    1. generate the .tmLanguage file
- `package`
    1. package the extension
- `install`
    1. install the extension

If no argument is passed then the help message is shown.

These operations can be chained (e.g. `./script/build build install`).

I've explored the possibility of using a makefile, but the YAML to plist script would still need to be run as a separate script, so we might as well keep the bash script.

# Fixes
Email patterns are now recognized inside other inline patterns (bold, italic, etc...)

Indentation in package.json has been "fixed" (i.e. `npm install` has changed it to follow its own guidelines). Only a few lines were changed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -339,12 +339,12 @@
 					"type": "boolean",
 					"default": true,
 					"description": "Use asciidoctor js instead of the 'asciidoctor_command'"
-                },
-                "asciidoc.forceUnixStyleSeparator": {
-                    "type": "boolean",
-                    "default": true,
-                    "description": "Force set the file separator style to unix style. If set false, separator style will follow the system style."
-                }
+				},
+				"asciidoc.forceUnixStyleSeparator": {
+					"type": "boolean",
+					"default": true,
+					"description": "Force set the file separator style to unix style. If set false, separator style will follow the system style."
+				}
 			}
 		},
 		"configurationDefaults": {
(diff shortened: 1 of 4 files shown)


joaompinto:
@MatteoCampinoti94 thanks once again.

Regarding the possibility of having a Makefile I would prefer not to have it, I would like to keep the extension as platform independent as possible, in that sense we should try to use nodejs and/or scripts defined at packages.json. Well Python is acceptable because I am a python lover and because it's easily installable on most platforms :P

MatteoCampinoti94:
@joaompinto No need :)

Well, bash is not available on Windows (unless you go through the Ubuntu for Windows thing), whereas CMake is cross-platform. That'd be the only reason really.
Hehe, I love Python too. The script requires the `yaml` package from pip but it should be available for all platforms.

Another idea could be using a single Python script. It'd be fully cross-platform and the `os.system()` function could be used to run the `npm` and `vsce` commands.

PR #140, opened 2018-12-04
 Use different css for preview, use editor colors for preview

MatteoCampinoti94:
Modified the `convert_using_javascript()` function in `text-parser.ts` to allow for an extra argument: `preview: boolean`.

When set to true the function loads a secondary .css named `ascidoctor-preview.css` instead of the default one. Default css is still used by `exportAsPDF.ts` (it simply passes `preview` as false).

The new `ascidoctor-preview.css` file has also been modified, removing the background and text colors from the body. First step towards using editor built-in colors for preview.


MatteoCampinoti94:
Nevermind, seems I managed to break it somehow Dx
Searching the problem now.

MatteoCampinoti94:
I have no idea what the problem is, for some reason it is using the right stylesheet variable but not loading it correctly.

MatteoCampinoti94:
Fixed! We were wrong, we didn't need a new stylesheet at all: ExportAsPDF already uses a different one (`all-centered.css`). We only have to modify `asciidoctor.css`.

MatteoCampinoti94:
I'm still not 100% sure this is fully true, but from what I could see from testing the output PDFs aren't touched by changes made to `ascidoctor.css`.

MatteoCampinoti94:
Okay, I've disabled all colors and backgrounds, except for specific classes like `.red`. This way the preview uses text and background taken from the editor, but can still render explicit color classes:
screenshot from 2018-12-05 02-45-04

Normal rendering seems to work correctly:
screenshot from 2018-12-05 02-46-46
screenshot from 2018-12-05 02-48-17
screenshot from 2018-12-05 02-48-44

Quotes have a weird shadow though:
screenshot from 2018-12-05 02-49-21

I'll look and see if I can find the right tag (or if it something due to the theme I'm using).

MatteoCampinoti94:
All disabled `color`, `background` and `shadow` attributes have simply been commented out with a specific scheme `/* DISABLED [attribute]: [value]; */` to be easy to find again and re-enabled if necessary.

Tags that would have resulted empty have been disabled instead of their attributes (i.e. `/* DISABLED [tag] { [attributes] } */`).

MatteoCampinoti94:
I've removed `font-family` attributes as well. Font styles (e.g. monospace) are left intact of course.

PR #147, opened 2018-12-11
Fix documents with apostrophe in path not rendering with manual command

MatteoCampinoti94:
As the title says, when using a custom asciidoctor command the rendering would fail if there was a non-closed apostrophe in the path (e.g. `path/to/document's`).

At first I tried solving it by replacing `'` with `\'` in the `documentPath` declaration but for some reason that didn't work. So instead I changed `'` with `"` to enclose the path in the system call (i.e. `'path/to/document'` -> `"path/to/document"`) and replaced `"` `\"` in `documentPath`. Not sure why this worked and the other version didn't.

### DIFF
--- a/src/text-parser.ts
+++ b/src/text-parser.ts
@@ -95,7 +95,7 @@ export class AsciiDocParser {
     private async convert_using_application(text: string) {
         const editor = vscode.window.activeTextEditor;
         const doc = editor.document;
-        const documentPath = path.dirname(this.filename);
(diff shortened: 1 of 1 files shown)


MatteoCampinoti94:
Added the full asciidoctor command to the error message. Used it to debug the apostrophe issue and thought it would be useful for others too.

The added line is:
`command: [command] [args] [document path]`

Issue #148, opened 2018-12-12
Make preview style easily configurable


mvuijlst:
While the issues mentioned in #139 are valid (it can be jarring to see a preview on a white background while editing in an otherwise dark environment, you're not necessarily interested in what the PDF version will look like, ...), there can be valid reasons to actually want to see the preview the way it used to be rendered. 

I am producing a file that will eventually end up as a document in black on white, I'd like to see it a little closer to what it will look like. What I see now is really not usable. 

![image](https://user-images.githubusercontent.com/4365702/49858560-88e00e00-fdf5-11e8-9b21-b42047934566.png)

I suggest making an option out of #139: either "use standard css for preview yes/no" or "follow VS code theme for preview yes/no".

I looked into the  `asciidoc.styles` configuration option, but as my project has many folders of asciidoc files and will be used in many enviroments (also off-line), it's just not a viable option.

MatteoCampinoti94:
Solved with #149.

A new option has been added asciidoc.preview.useEditorStyle of type bool:
* `true`
the modified .css is used which partially inherits the editor style.
* `false`
original .css is used instead.

Furthermore setting `asciidoc.use_asciidoctor_js` to `false` forces the extension to use the system-installed `asciidoctor` command which uses its own stylesheet (the default one).

PR #149, opened 2018-12-12
Add option to use editor or default style in preview

MatteoCampinoti94:
As per the title, I'm trying to address #148.

A new option has been added `asciidoc.preview.useEditorStyle` of type bool.
If set to true the modified .css is used which partially inherits the editor style.
If set to false original .css is used instead.

For the moment the setting needs a window reload to be applied (i.e. `Ctrl+Shift+P` `Reload Window`) if changed. I'll work more on it to try and fix this issue.


MatteoCampinoti94:
Fixed it. The extensions now loads both preview styles when it starts and then chooses which one to use depending on the new setting.

PR #158, opened 2019-01-17
To fix a problem when preview multiple files in multiple folders with plantuml

Dimeiza:
Hello,
I am always helped by this useful tool.

Problem:

When open multiple asciidoc files with separated folders, and use with plantuml in indivisual asciidoc files, all diagram images output to same folder.
And preview of one file don't output plantuml image.

Screenshot:
preview_folder_a

preview_folder_b

Cause and pull request:

It seems that setting of documentpath  (in convert_using_application) is a cause of this problem.
After fixed it, works fine.

preview_folder_a

preview_folder_b

Could you check this pull request? 

Thank you for taking your time.

### DIFF
--- a/src/text-parser.ts
+++ b/src/text-parser.ts
@@ -95,7 +95,7 @@ export class AsciiDocParser {
     private async convert_using_application(text: string) {
         const editor = vscode.window.activeTextEditor;
         const doc = editor.document;
-        const documentPath = path.dirname(this.filename).replace('"', '\\"');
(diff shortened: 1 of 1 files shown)


MatteoCampinoti94:
Have you checked if it causes any issues in other situations?

Dimeiza:
Although I have already used it within a few days, I haven't found any problems.

When I previewed sample files this repository included, I seemed that there is no problem.

Please let me know if there are additional test samples for this patch. I will do it.

Issue #167, opened 2019-03-05
jsonValidation setting causing unfixable problem to appear in Problems panel


daogilvie:
Hi there!
I'm on vscode-insiders (Version 1.32.0-insider (1.32.0-insider)), and I noticed that the `jsonValidation` setting in your `package.json` appears to trigger an ENOENT warning every time I open any `package.json` in any project, as it points to a schema file that does not exist. Removing the entry from the asciidoctor-vscode `package.json` in my extensions directory makes the error go away, and doesn't seem to harm the functioning of the extension. I think this error has been happening to me for some time, but this is the first time I've had the time to investigate it. Would you object to removing the setting? Or including the schema file in the repo so the reference URL is valid? I tried to find VScode guidelines for disabling the schemas in extensions but I haven't found too much yet. Thanks for maintaining this, I'm trying to get my office to see the value of Asciidoc and having this extension is really useful 👍

MatteoCampinoti94:
Closed with pull request #168 (commit 629ea66).

PR #168, opened 2019-03-06
remove jsonValidation section of package.json

shaneknysh:
jsonValidation references a folder that is not included in the install


shaneknysh:
fixes #167

PR #178, opened 2019-03-27
resolves #175 set RUBYOPT='-E UTF-8' when invoking the asciidoctor command

### LINKED ISSUE #175: Mojibake unstead of cyrillic
Hello.

vscode 1.32.3 on Windows 8
asciidoc v2.3.3

I **don't** use `asciidoctor.js`. I use Ruby version of Asciidoctor. But when I tried `asciidoctor.js`, all worked fine.

image

I tried Chrome extensions. It works without any problems.

I tried in console `asciidoctor test.adoc` &mdash; all works fine too.

```
asciidoctor -v
Asciidoctor 2.0.1 [https://asciidoctor.org]
Runtime Environment (ruby 2.4.4p296 (2018-03-28 revision 63013) [x64-mingw32]) (lc:IBM866 fs:Windows-1251 in:IBM866 ex:IBM866)
```

-- Andrey.

mojavelinux:
This change is required for the `asciidoctor` command in Asciidoctor 2.0 to work properly on a system that configures Ruby to use an IO encoding other than UTF-8.

### DIFF
--- a/src/text-parser.ts
+++ b/src/text-parser.ts
@@ -100,7 +100,7 @@ export class AsciiDocParser {
 
         return new Promise<string>(resolve => {
             let asciidoctor_command = vscode.workspace.getConfiguration('asciidoc').get('asciidoctor_command', 'asciidoctor');
-            var options = { shell: true, cwd: path.dirname(this.filename) }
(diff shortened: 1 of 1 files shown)


mojavelinux:
You could take this a little further and respect an existing RUBYOPT value. I'll submit another PR for you to consider.

PR #179, opened 2019-03-27
merge with existing RUBYOPT environment variable

mojavelinux:


mojavelinux:
..since the RUBYOPT environment variable may contain other settings. We just need to override the `-E` option.

PR #196, opened 2019-05-09
Feature/better image pasting windows

jacksoncougar:
The current version on vscode doesn't support pasting images.

I've implemented a lightweight paste that will work on windows OS using powerscript
- supports saving images to the current folder pointed to by the `imagesdir` attribute
- supports using selected text as the filename of the images
- supports inline- and block-macros depending on the context when pasting images 
- automatically adjusts whitespace of inserted image macro to keep macro valid


MatteoCampinoti94:
That's great! I'm travelling today so it'll take me a bit, but as soon as I can I'll check the new code.

MatteoCampinoti94:
I've looked it over and the code seems to work really well, but I want to wait a bit before pulling it into our master and see if I can add multi-platform functionality. We are planning an overhaul of the base code and we might directly pull some of the changes in the new version.

Thank you again for contributing!

PR #240, opened 2019-12-15
Add code to ensure SymbolProvider not called too frequently,

### LINKED ISSUE #238: making VSCode very slow
When I have this extension enabled, VSCode becomes very slow to respond to key strokes.
Simply disabling the extension returns responsiveness of the UI back to normal.

## System Information

Version: 1.40.2
Commit: f359dd69833dd8800b54d458f6d37ab7c78df520
Date: 2019-11-25T14:52:45.129Z
Electron: 6.1.5
Chrome: 76.0.3809.146
Node.js: 12.4.0
V8: 7.6.303.31-electron.0
OS: Darwin x64 19.0.0

AsciiDoc 2.7.7

## To Reproduce

Steps to reproduce the issue:

1. Enable AsciiDoc
2. Try to type in a .adoc file
3. Notice slow response to keystrokes.
4. Disable AsciiDoc
5. Type in the .adoc file again.
6. Notice that response to keystrokes is fine now.

danyill:
Closes #238 

I'm not too familiar with TypeScript interfaces so I've added several null arguments to allow the new class variables to retained.

I'd be very pleased for a review and happy to make any changes.

*What this does:*

Previously on some machines it was found that this extension with the recently added symbol/outline view (in v 2.7.7, #234) would cause them to go very slowly (see #238).

This seems to be because the new outline view uses Asciidoctor.js to produce the outline and takes the sections from the AST. VS code fires `provideDocumentSymbols` frequently and if the processing time from Asciidoctor.js is slow than the calling rate from VS code then a queue of asynchronous `provideDocumentSymbols` start to grow. This can then use significant CPU capacity and starve other processes as the user types. The queue then clears slowly.

I've changed the way the extension runs so that initially on load it produces an outline. It then records the processing time to produce the outline and will not produce an outline more frequently than `RunTimeFactor`, here set to 3.

I've tested this on a couple of Linux computers using the following version of code:

```
1.41.0
9579eda04fdb3a9bba2750f15193e5fafe16b959
x64
```

I find that the CPU goes down almost immediately after ceasing typing which is the desired effect. I have not metricated this using the debugger tools in code.

It would be good to test this on a variety of documents before we merge.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -33,7 +33,7 @@ export function activate(context: vscode.ExtensionContext) {
 	];
 
 	const contentProvider = new AsciidocContentProvider(engine, context, cspArbiter, contributions, logger);
-	const symbolProvider = new AdocDocumentSymbolProvider(engine);
+	const symbolProvider = new AdocDocumentSymbolProvider(engine, null, null, null, null);
     const previewManager = new AsciidocPreviewManager(contentProvider, logger, contributions);
 	context.subscriptions.push(previewManager);
 
--- a/src/features/documentSymbolProvider.ts
+++ b/src/features/documentSymbolProvider.ts
@@ -24,7 +24,7 @@ export default class AdocDocumentSymbolProvider implements vscode.DocumentSymbol
 		},
 		private lastSymbolCall: number,
 		private lastRunTime: number = 1000,
-		private RunTimeFactor: number = 5
+		private RunTimeFactor: number = 1.5
 	) { }
 
 	public async provideDocumentSymbolInformation(document: SkinnyTextDocument): Promise<vscode.SymbolInformation[]> {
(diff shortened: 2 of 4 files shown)


danyill:
I've done a little more testing. For files of say 10klines and >300 symbols I'm still seeing excessive CPU consumption and symbol provision doesn't end. Running the asciidoctor.js runtime against these they may take 2-4 seconds to run. Running the extension in debug mode, these can take a **very long time** to process (1 min, 5 mins, 10 mins). 

This is very greatly improved with #241 because they pull from the same parsing code in the `foldingProvider`. I've added these commits here as well. In this case processing might take 10-15 seconds on first run and the symbols update much more slowly (I have a factor of 5 in the code, so symbols would update in 50-75 seconds). Probably 5 is too conservative but 1 is potentially too low under variable CPU load.

mosabua:
Could we get this looked at, merged and released. Currently the asciidoc extensions is impossible for me to use due to the CPU load generated

Issue #239, opened 2019-12-01
Export to pdf is not rendering math expressions and pictures


WyRe:
## Description

When you use the export to pdf option math expressions and pictures are not rendered.

## System Information

OS: Arch Linux 5.3.13 kernel version
VSCode version: 1.40.2
Asciidoctor extension: v2.7.7

## To Reproduce

Steps to reproduce the issue:

1. Create an `.adoc` example document with a MWE like this one
```
:stem: latexmath

= Table example with latex syntax

== Testing

I'm also testing a few admonitions

[NOTE]
.First admonition
Here you have a note. 
====
.Table inside a block
[cols=".^,<.^a,>.^a"]
|===
| (Test) \[X_nY_m\]
| * \(X\) with \(i\)
* \(Y\) with \(j\)
| [cols=">.^"]
!===
! FOO ! BAR
!===
|===
====

IMPORTANT: Here a second admonition with no title.

.Image
image::./picture.jpg[Picture!]
```
2. Use `ctrl+shift+p` to open the command palette and choose `AciiDoc: Export document as PDF`
3. Open the generated pdf file, you should not see properly rendered the math expressions an the picture.

## Screenshots & Files

![Captura de pantalla_2019-12-01_13-26-08](https://user-images.githubusercontent.com/4059108/69913915-318be100-143e-11ea-914c-0eeb5019190f.png)

danyill:
I think there is two unrelated things here.

The lack of embedding images is an upstream issue, see https://github.com/wkhtmltopdf/wkhtmltopdf/issues/2687. To work around this I believe if you set `data-uri` on your document by including `:data-uri:` in the document header you will see images (I do in my testing). This embeds the data directly into the html file which is handled better by wkhtmltopdf.

The lack of equations seems to be a problem with SSL negotiation in wkhtmltopdf. I could reproduce this issue in `wkhtmltopdf 0.12.4 (with patched qt)` but not when I symlinked it to a local install of `wkhtmltopdf 0.12.5`. 

@joaompinto can we update to 0.12.5? I'm not sure what process we use for retrieving and updating these binaries.

Note that the other option is to use `asciidoctor-pdf` (the Ruby version) in conjunction with e.g. `asciidoctor-mathematical` which will render equations and embed images.
Within a short time I think we will be able to support `asciidoctor-pdf`` (the Javascript version) which is similar to wkhtmltopdf (but, I think likely better for Asciidoc users).

[...]

PR #252, opened 2020-01-28
Prepare for compatibility with asciidoctor-pdf.js, related to #239

danyill:
I'd like this extension to be able to support asciidoctor-pdf.js 

Support for reading from stdin was recently added but the @asciidoctor/cli requires that the stdin argument be the last one. So this PR simply moves this argument to the end which has no impact on asciidoctor-pdf (Ruby).

See https://github.com/Mogztter/asciidoctor-pdf.js/issues/155 for details.


joaompinto:
LGTM

PR #253, opened 2020-01-31
Add classes from <body> in Asciidoctor output to webview document

### LINKED ISSUE #141: Preview renders incorrectly with toc:left
The table of contents is incorrectly rendered over the text when toc is set to left when your window is not wide enough. 

When asciidoctor renders to html, the toc gets a width of 15 or 20em, and the body a corresponding padding-left of 15 or 20em. 

This does no longer seem to happen in asciidoctor-vscode. 

Correct:

correct

Incorrect:

incorrect1

incorrect2

danyill:
Closes #141 

The core problem is that classes on the `<body>` element are missing which means the default asciidoctor stylesheet is not functioning correctly resulting in the sidebars associated with the table of contents not presenting as user's expect (being on top of the text).

A normal Asciidoctor document with `:toc: left` might look like this:

image

The way we process Asciidoctor content is by rendering it from the lines of the source document and then fitting it into a template expression. Probably we should be using the embedded option (see [here](https://asciidoctor.org/docs/user-manual/#using-a-toc-with-embeddable-html)) and then handling the table of contents in a different fashion (by templating it and styling it specific to this extension).

I think (but am not sure) that the `<body>` element of the Asciidoctor output is being removed (not sure by what) so that only the top-level (actually second level) `<body>` element exists. I'm not sure if this is done in this extension or by the VS Code html parser (I think this is invalid html "strictly" speaking). This is then having the unintended affect of removing the classes necessary to make the table of contents function correctly.

So the DOM for the preview output looks like:

image

Note that there is two `<body>` tags, the second in the `<iframe>` but the third (!) `<body>` tag provided by L#70 and L#85 is missing in the output.

https://github.com/asciidoctor/asciidoctor-vscode/blob/db4a1d125fae2b22552d5c1056848213162781e7/src/features/previewContentProvider.ts#L70-L90

This PR provides a somewhat "hacky" fix which is to use a (hopefully narrowly scoped) regex to pick up the Asciidoctor output `<body>` classes and put them into the output html. I have done this because otherwise we would need to parse into a DOM for which we currently don't require any libraries and it would take significant time. However using a regex for this is very likely brittle and may cause other problems.

I've tested this with `:toc: right|left|auto|macro` and it appears to work, but I'd appreciate others carrying out some tests.

Happy to work on a better solution if someone can provide a few suggestions on an alternative approach.

### DIFF
--- a/src/features/previewContentProvider.ts
+++ b/src/features/previewContentProvider.ts
@@ -68,6 +68,9 @@ export class AsciidocContentProvider {
 		const nonce = new Date().getTime() + '' + new Date().getMilliseconds();
 		const csp = this.getCspForResource(sourceUri, nonce);
 		const body = await this.engine.render(sourceUri, config.previewFrontMatter === 'hide', asciidocDocument.getText());
+
+		const bodyClassesRegex = /<body(?:(?:\s+class(?:\s*=\s*(?:\"(.+?)\"|\'(.+?)\')))+\s*)>/
+		const bodyClasses = body.match(bodyClassesRegex)
 		return `<!DOCTYPE html>
 			<html>
 			<head>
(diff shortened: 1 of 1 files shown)


joaompinto:
I can't think of a better option right now, merging it as is, will do some tests over the weekend.

PR #254, opened 2020-01-31
Use safe_load to avoid warning

danyill:
This PR just uses `safe_load` to avoid the error message in the script.

```bash
10:09 $ ./build.sh generate
==> Converting YAML syntax to plist
script/yaml-to-plist.py:30: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  out_file.writelines(convert(yaml.load(in_file)))
```

I've diffed the output with either method and both methods convert identically.


joaompinto:
LGTM

PR #256, opened 2020-01-31
WIP: Update dependencies generally, move to Asciidoctor.js 2.1.0 and Asciidoctor 2.0.10

danyill:
Update various dependencies -- I don't think we have tests so it would be good if someone else other than me tested the features they care about -- but I don't see any breakages on my machine (Linux, Kubuntu).

Also quite a big version bump for asciidoctor-plantuml.

Perhaps we could leave this as WIP for a little while while I do some testing.


joaompinto:
Until we bundle some tests or VSCode provides some kind of branching for extensions the individual testing is always rather limited. I merge this commit, will do a release today, and will have some extra attention for bug reports.

Thanks

PR #258, opened 2020-02-01
Clear out old roadmaps, merge old CHANGELOG, mention recent merges/issues

danyill:
This is just a little tidy up/release prep:

* remove old ROADMAP
* integrate old CHANGELOG and remove
* update and document author list


README.md
```diff
@@ -172,15 +179,15 @@ All the following people who have contributed to the extension:
 * Johannes Rössel
 * Joshua Stafman
 * Kevin Palmowski
-* Loïc PÉRON
+* Loïc PÉRON
```
waldyrious: The characters with diacritics in this line were changed from single characters to [composite characters](https://en.wikipedia.org/wiki/Precomposed_character):
- `ï` (U+00EF) became `ï` (U+0069, U+0308)
- `É` (U+00C9) became `É` (U+0045, U+0301)

I assume this change was unintended, so it should probably be reversed.

CHANGELOG.md
```diff
@@ -106,3 +119,145 @@
 * YAML syntax file, improved inline syntax detection, automated building (thanks to matteo.campinoti94@gmail.com)
 * fixes typos in Description for AsciiDoc.forceUnixStyleSeparator (thanks to ccb012100)
 * diagram (specifically plantUML) support in asciidoctor.js (closes #105)
+
+# Changes
```
waldyrious: Was this heading left here on purpose? Seems like an overlook.

joaompinto: I will fix after merging, thanks for the look

README.md
joaompinto: If correctly understand this change the chars have been changed to the UTF-8 encoding, should be fine

waldyrious: Hmm, I'm not sure that's the case. From what I could ascertain, `ï` (U+00EF) is encoded in UTF-8 as the byte sequence `C3 AF`, whereas the composite `i`+`¨` is `69 CC 88`; and similarly, `É` (U+00C9), is encoded in UTF-8 as `C3 89`, whereas the composite `E`+`´` is encoded as `45 CC 81`.

The new part of the diff uses the composed characters, so something other than encoding conversion seems to have caused the issue.

waldyrious: I've opened #259 to restore the original non-composed characters in this line.

PR #259, opened 2020-02-02
Fix Unicode characters in README

waldyrious:
(See the conversation [here](https://github.com/asciidoctor/asciidoctor-vscode/pull/258#discussion_r373777196) for context.)

Two characters with diacritics in the README were changed in 6f387f0 from single characters to composite characters:

- ï (U+00EF = `C3 AF` in utf-8) became ï (U+0069 U+0308 = `69 CC 88` in utf-8)
- É (U+00C9 = `C3 89` in utf-8) became É (U+0045 U+0301 = `45 CC 81` in utf-8)

Note that the change did not affect other similar characters, such as `ö` (U+00F6), which remained in their non-composed forms.

This commit restores the two changed characters to the original ones.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,7 +179,7 @@ All the following people who have contributed to the extension:
 * Johannes Rössel
 * Joshua Stafman
 * Kevin Palmowski
-* Loïc PÉRON
+* Loïc PÉRON
 * Marcelo Alvim
 * Mark Roszko
 * Masanori Asano


danyill:
The change was caused by running the following command on this repository to update the author list.

```
git log --format='%aN' | sort -u | awk '{print "* "$0}' | uniq
```

and dumping the output into the README.

I manually disambiguated that author (and one or two others) who had used multiple ways to spell their names.

I really have no particular point of view as to what is right or wrong. 

I'm somewhat included to think that this doesn't really matter unless either (a) there's a canonical way this should be done or (b) you're the person referred to and you have a strong preference for which unicode glyphs make up your name...

joaompinto:
@danyill I agree with you that in general it doesn't matter, nevertheless I will accept this fix from  @waldyrious just because it mattered to him enough to submit the PR :)

waldyrious:
Thanks for the clarifications and for indulging me :)

> I'm somewhat included to think that this doesn't really matter unless either (a) there's a canonical way this should be done 

That is indeed the case! 🙂 From the [Wikipedia article on composite characters](https://en.wikipedia.org/wiki/Precomposed_character) (emphasis mine):

> Precomposed characters are the **legacy solution** for representing many special letters in various character sets. **In Unicode they are included primarily to aid computer systems with incomplete Unicode support**, where equivalent decomposed characters may render incorrectly.

PR #260, opened 2020-02-02
Support image pasting on Linux and Mac

### LINKED ISSUE #255: Paste image on linux fails
## Description

When pasting images on Linux (Kubuntu 19.10) I receive an error message and no image appears.

## System Information

Version: 1.41.1
Commit: 26076a4de974ead31f97692a0d32f90d735645c0
Date: 2019-12-18T15:04:31.999Z
Electron: 6.1.5
Chrome: 76.0.3809.146
Node.js: 12.4.0
V8: 7.6.303.31-electron.0
OS: Linux x64 5.3.0-24-generic

## To Reproduce

On Linux use the Paste Image command

## Screenshots & Files

image

danyill:
This closes #255 by adding paste image support for Linux.

Sorry @joaompinto hope you haven't started significant work on this yet.

I based my initial work on https://github.com/mushanshitiancai/vscode-paste-image/blob/master/src/extension.ts

It should also allow image pasting for MacOS but I have no Mac hardware to test on and there are very likely problems.

While doing this I made a few other improvements (at least I think they are):

* If regex parsing of the document doesn't yield `imagesdir` then process the document retrieve from the AST.
* If the user enters a filename don't require them to append `.png` to the image, but have the extension carry this out.
* If a line begins with non space separated text, paste as a block image not an inline image
* The previous way paths were derived didn't look quite right to me so I've modified to it work off the current document's `filePath` not the workspace path.

Would appreciate any testing


joaompinto:
your work came in time, one less for me :) LGTM

joaompinto:
The paste image function with an empty doc and no image actually copied resulted in:

TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received type undefined

Will check this before release.

PR #265, opened 2020-02-08
Add shortcut key for paste image, clean up whitespace, ensure empty d…

danyill:
Add shortcut key for paste image, clean up whitespace, ensure empty document image save error message is clear

@joaompinto thanks for testing, this resolves your comment in https://github.com/asciidoctor/asciidoctor-vscode/pull/260#issuecomment-581779947

* tidies up unused references
* removes whitespace
* Adds a shortcut key for linux/mac (same as https://marketplace.visualstudio.com/items?itemName=mushan.vscode-paste-image)

Finally the prediction path for inline vs block images made no sense to me. Whether we are replacing or inserting text, if there is already text on a line (other than, say spaces) and other than a single "word" (which we need for when the user has provided a name for the image, we don't allow images with spaces) we should use the single `:` (inline image) otherwise we should use the block image (`::`). At least insofar as I understand it.

```
some text <selection "asdf"> ==> image:[]
<selection "one_word_image_filename"> ==> image::[]
```

On my machine for reasons which are unclear when debugging I get a JSON error with `package.json`. But I believe it to be valid JSON (web tools seem to agree) and the package builds properly. Perhaps you might know why this happens or be able to confirm this problem?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Change language definition to be the same as that of the Atom extension [atom-language-asciidoc](https://github.com/asciidoctor/atom-language-asciidoc/)
 * Support the source block (closes #212)
 * Support image pasting on Linux/Mac (closes #255)
+* Add Ctrl+Alt+V/Cmd+Alt+V (Mac) as shortcut key for image paste
 * Add `.png` extension automatically for image pasting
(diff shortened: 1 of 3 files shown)


joaompinto:
LGTM

PR #266, opened 2020-02-11
Add command to save the document HTML

danyill:
This PR adds a new command to save the document to HTML, normally using the default stylesheet unless the user has specified another (as opposed to the editor preview stylesheet).

Hopefully this is a useful addition. As always, happy to receive reviews/improve where requested.

I've updated the engine to support `vscode.env.openExternal` and done the migration to the newer package as per https://code.visualstudio.com/api/working-with-extensions/testing-extension#migrating-from-vscode (not for tests but for the rest).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.7.11
+
+* Add support for saving html output
+
 ## 2.7.10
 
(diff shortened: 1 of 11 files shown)


danyill:
I have added icons for this command but have not done anything with them -- they may be of some use in the future.

PR #270, opened 2020-02-16
Add Save to Docbook option, closes #102

### LINKED ISSUE #102: Add export to HTML and DocBook
It would be neat to have a Command Palette option to export created Asciidoc document to HTML and DocBook formats supported by AsciiDoctor.

danyill:
* Adds Docbook support, closes #102
* Also rearranged arguments for command line options to support command line asciidoctor.js better (stdout/stdin must be last).

Would appreciate some testing, particularly on Windows if possible.

I don't really use Docbook so after we're comfortable this will basically work and roll it into a release it would be good to ask on the issue for people to test/comment.


joaompinto:
LGTM 
I will test it on Windows

PR #276, opened 2020-03-01
Update webpack configuration to improve file bundling

danyill:
When building this package there is usually a message given:

> This extension consists of 2726 files, out of which 2071 are JavaScript files. For performance reasons, you should bundle your extension: https://aka.ms/vscode-bundle-extension . You should also exclude unnecessary files by adding them to your .vscodeignore: https://aka.ms/vscode-vscodeignore

I have followed the instructions from:

https://code.visualstudio.com/api/working-with-extensions/bundling-extension

and tried to make this extension conform with them.

Noting:

* we shouldn't bundle all of `node_modules` as this isn't good for performance.
* I think the only thing we need to exclude from webpack is the Asciidoctor opal version which webpack produces errors if I include it (now excluded in `webpack.config.js`)

Before: 

```bash
19:02 $ ./build.sh package
==> Packaging extension
Executing prepublish script 'npm run vscode:prepublish'...

> asciidoctor-vscode@2.7.13 vscode:prepublish /home/mulhollandd/Documents/asciidoctor-vscode
> tsc -p ./ && npm run build-preview


> asciidoctor-vscode@2.7.13 build-preview /home/mulhollandd/Documents/asciidoctor-vscode
> webpack --mode development

Hash: 287e6d0a4c7ccad2789b
Version: webpack 4.41.6
Time: 2038ms
Built at: 03/01/2020 7:02:55 PM
   Asset      Size  Chunks             Chunk Names
index.js  88.2 KiB   index  [emitted]  index
  pre.js  27.6 KiB     pre  [emitted]  pre
Entrypoint index = index.js
Entrypoint pre = pre.js
[./node_modules/webpack/buildin/global.js] (webpack)/buildin/global.js 472 bytes {index} [built]
[./preview-src/activeLineMarker.ts] 1.11 KiB {index} [built]
[./preview-src/csp.ts] 2.04 KiB {pre} [built]
[./preview-src/events.ts] 659 bytes {index} [built]
[./preview-src/index.ts] 4.5 KiB {index} [built]
[./preview-src/loading.ts] 1.25 KiB {pre} [built]
[./preview-src/messaging.ts] 682 bytes {index} [built]
[./preview-src/pre.ts] 550 bytes {pre} [built]
[./preview-src/scroll-sync.ts] 4.68 KiB {index} [built]
[./preview-src/settings.ts] 1010 bytes {index} {pre} [built]
[./preview-src/strings.ts] 696 bytes {pre} [built]
    + 1 hidden module
[...]

PR #311, opened 2020-05-12
WIP: Remove asciidoctor-plantuml, closes #308

### LINKED ISSUE #308: Remove asciidoctor-plantuml
asciidoctor-plantuml is [no longer maintained](https://github.com/eshepelyuk/asciidoctor-plantuml.js
).

Fortunately thanks to a recent contribution we already have the recommended alternative which is using [asciidoctor-kroki](https://github.com/Mogztter/asciidoctor-kroki).

It will simplify this extension and provide a better experience to use only asciidoctor-kroki in my view. Asciidoctor-kroki also provides a free public instance.

@joaompinto Any thoughts? Any objections to me submitting a PR for this?

danyill:
This PR removes asciidoctor-plantuml support as this extension is no longer maintained and asciidoctor-kroki is the suggested replacement.

I'd like to wait a little while before merging it until a couple of open PRs  in asciidoctor-kroki (https://github.com/Mogztter/asciidoctor-kroki/pull/69, https://github.com/Mogztter/asciidoctor-kroki/pull/64) are addressed to minimise end user disruption (asciidoctor-kroki is not an identical replacement to asciidoctor-plantuml yet but these two PRs will help).


danyill:
A new version of asciidoctor-kroki has been released (v0.9.1)
I've updated and done some testing on this and all appears to be good and so I've integrated it.

However I've recently rebuilt my machine and noticed more problems with the webpack-ing of `node_modules` (see #276). This was quite new to me when I did it but I expect I had system installs of asciidoctor.js and so my NODE_PATH was covering the deficiency in the built package. On my rebuilt  machine which is fairly clean I noticed that dependencies of `asciidoctor-opal-runtime` (and their dependencies etc.) do not get webpacked correctly. For now I've just included them manually (in `.vscodeignore` and the webpack configuration) as I'm not sure exactly why this is (any help appreciated).

@anb0s has put some upstream effort into plantuml includes and is keen to have this support in this extension.

@joaompinto could we do a release soonish please :pray: 
I wish there was a pre-release feature so that we could get more orderly feedback

I would appreciate anyone who can doing some testing because I worry about the webpack MR which has not yet been part of a release and could break functionality (when time allows I must look into a better testing framework because we have many users and it's currently too easy to break things with infrastructure changes).

Currently the major gap (I think) between this and asciidoctor-plantuml is the ability to name files (https://github.com/Mogztter/asciidoctor-kroki/issues/90) and some options. As a result existing users relying on asciidoctor-plantuml behaviour may notice some differences.

joaompinto:
@danyill I noted now that the scroll sync in master is broken, I will try to identify and fix it, I will do a release once the scrlol sync is fixed.

joaompinto:
Release done

PR #326, opened 2020-06-10
Added the ability to use the workspace path as the base_dir

cirrusj:
This PR adds a configuration option that allows to use the workspace path (when using a workspace) as the asciidoctor base_dir.
This option will be very useful when having a multi-directory documentation project, that includes files from a common directory. For example:
```
project/
├── common
│   └── settings.adoc
├── data
│   └── file1.adoc
└── index.adoc
```
Where file1.adoc uses: `include::{docdir}/common/settings.adoc[]`.
Previewing file1.adoc would display an `Unresolved directive in <stdin> - include::{docdir}/common/settings.adoc[]` message.

### DIFF
--- a/package.json
+++ b/package.json
@@ -420,6 +420,11 @@
 						"%asciidoc.saveDocbook.docbookVersion.docbook5%",
 						"%asciidoc.saveDocbook.docbookVersion.docbook45%"
 					]
+				},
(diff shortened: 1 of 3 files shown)


joaompinto:
LGTM, thanks for the contribution

PR #350, opened 2020-07-12
Fix unexpected splitting multibyte chars while chunking

nunulk:
when "save HTML document" command executed, it often happens a multibyte character is unexpectedly split by Stream chunking like the following.

```javascript
{ data: '...コンピュータプログラムの明�' }
{ data: '��化を目的にし...' }
```

refs: https://nodejs.org/api/child_process.html#child_process_maxbuffer_and_unicode

it can be the solution to replace string concatenation with Buffer's one and convert to string when the stream closed.


joaompinto:
Thanks for the fix

nunulk:
@joaompinto 
that's my line. thanks for merging.

Issue #377, opened 2020-11-17
Autocompletion for image:: and include:: files


rdmueller:
The IntelliJ-Asciidoctor-Plugin manages to do an autocomplete on `image::` and `include::` macros to select a file.

This is quite helpful and avoids typos. I would like to see this feature here in the asciidoctor-vscode plugin :-)

Details:

* You start to type `include::` and press `ctrl`+`space`. 
* a drop down menu opens with a list of current folders and files to select from
* you select a folder and the next drop down opens, until you select a file
* you close your `include::` macro by adding `[]`

d-koppenhagen:
@rdmueller should it really be necessary to press a keyboard shortcut like `ctrl` + `space`? I thing it would be even more easy and intuitive if the completion starts directly after typing `include::`.
This could work just like in the **PathIntellisense** Plugin (see here: https://github.com/ChristianKohler/PathIntellisense). WHat do you think about it?

rdmueller:
sure! even better if you don't have to activate it 😀

d-koppenhagen:
@rdmueller in addition, I added autocompletion also when using with defined variables. The completion will take tthe he pathes defined in the varaible selected from the completion list and starts to search for child elements inside the directory

rdmueller:
wow! sounds perfect!
And do I see a PR there? awesome!

d-koppenhagen:
Yes. There is still a small problem with just a specific case I linked a stackoverflow issue. Currently I have no idea how to solve it

eiswind:
is there any progress on this? atm it seems a little broken to me.

Mogztter:
> is there any progress on this? atm it seems a little broken to me.

Could you please be more specific?
Maybe you are referring to the fact that built-in attributes are included in the auto-completion?

This has been fixed in https://github.com/asciidoctor/asciidoctor-vscode/pull/529 and will be available in the next release.
Here's what I get using the master branch (next release):

![autocomplete](https://user-images.githubusercontent.com/333276/157022144-f0f3e587-8994-4910-af49-7d62bf27bc62.gif)

Mogztter:
@eiswind Thanks that's really valuable. I think we should open a new issue to track these two feature enhancements. 

> I found that the completion seems to resolve the files and folders to the root of the project, which makes it almost useless

[...]

PR #381, opened 2020-11-27
feat: provide import completion for include and images

d-koppenhagen:
related to #377

### DIFF
--- a/src/providers/asciidoc.provider.ts
+++ b/src/providers/asciidoc.provider.ts
@@ -38,8 +38,19 @@ function shouldProvide(context: Context): boolean {
 async function provide(
   context: Context
 ): Promise<vscode.CompletionItem[]> {
+  const documentText = context.document.getText();
   const pathExtractedFromIncludeString = context.textFullLine.replace('include::', '').replace('image::', '').replace('image:', '');
-  const entryDir = pathExtractedFromIncludeString.substr(0, pathExtractedFromIncludeString.lastIndexOf("/"));
+  let entryDir = pathExtractedFromIncludeString.substr(0, pathExtractedFromIncludeString.lastIndexOf("/"));
+
+  // use path defined in a variable used
+  if (entryDir.startsWith('{')) {
+    const variableName = entryDir.replace('{', '').replace('}', '')
+    const match = documentText.match(new RegExp(`:${variableName}:.*`, 'g'));
+    if (match && match[0]) {
+      entryDir = match[0].replace(`:${variableName}: `, '');
+    }
+  }
+
   const workspace = vscode.workspace.getWorkspaceFolder(context.document.uri);
   const rootPath = workspace?.uri.fsPath;
   const searchPath = getPathOfFolderToLookupFiles(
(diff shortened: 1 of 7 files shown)


d-koppenhagen:
still WIP

d-koppenhagen:
`editor.autoClosingBrackets` makes some problems I need to solve: https://stackoverflow.com/questions/65079743/prevent-autoclosingbrackets-when-inserting-content-as-result-of-registercompleti

src/providers/asciidoc.provider.ts
```diff
@@ -0,0 +1,97 @@
...
+    kind: vscode.CompletionItemKind.Folder,
+    sortText: '..',
+  }
+  const globalVariableDefinitions = documentText.match(/:\S+:.*/g)
+
+  let variablePathSubstitutions = []
+  // TODO: prevent editor.autoClosingBrackets at this point until finished inserting
+  const editorConfig = vscode.workspace.getConfiguration('editor');
+  const doAutoCloseBrackets = editorConfig.get('autoClosingBrackets') === 'always';
+  if (globalVariableDefinitions) {
```
danyill, line 76: We already have autocompletion on the brace character `{` from the Asciidoc AST (see https://github.com/asciidoctor/asciidoctor-vscode/blob/master/src/features/attributeCompleter.ts).

src/providers/asciidoc.provider.ts
```diff
@@ -0,0 +1,97 @@
...
+}
+
+/**
+ * Provide Completion Items
+ */
+async function provide(
+  context: Context
+): Promise<vscode.CompletionItem[]> {
+  const documentText = context.document.getText();
+  const pathExtractedFromIncludeString = context.textFullLine.replace('include::', '').replace('image::', '').replace('image:', '');
```
danyill, line 42: Ideally we would resolve `image::` references in the completion relative to the document base directory and the `imagesdir` if set.

danyill:
Thanks for this great contribution, I've left a few comments if you're interested in examining a bit further a few small aspects but otherwise it looks good to me.

src/providers/asciidoc.provider.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  return [
+    levelUpCompletionItem,
+    ...variablePathSubstitutions,
+    ...items.map((child) => {
+      const result = createPathCompletionItem(child);
+      result.insertText = result.kind === vscode.CompletionItemKind.File ? child.file + '[]' : child.file
+      return result;
+    }),
+  ];
```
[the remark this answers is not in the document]
d-koppenhagen: solved here 
ee69fe2

joaompinto:
Thanks for the contribution, looks good to me.

joaompinto:
I have tested the include:: autocomplete, it does not provide me a list of the files, screensot:
autocomplete_issue

Did I miss something ?

d-koppenhagen:
Not sure but when I check out the master branch now and run the extension locally it seems to work fine.
Did you miss to scroll down? probably the suggestion order need to be re-thinked but I am not sure what will appear first in the list as other extension may contribute also to the list.

Here is a screenshot of my local test workspace:

Screenshot 2021-01-23 at 13 10 30

eiswind:
I found that in 2.9.5 it doesn't work when I have attributes in the include path.
include::{srcdir}/002_config/src/main/java/de/eiswind/training/spring/config/Application.java

another thing I noticed is that the base path is the project directory. Shouldn't inlcudes be relative to the adoc file?

Any help would be apprieciated.

Issue #361, opened 2020-08-27
plantUML diagrams included with makro syntax fail to render in the preview when they include further files


jacques-ro:
### Environment

![image](https://user-images.githubusercontent.com/61683053/91448803-a3347680-e87a-11ea-92e9-c650cbcd6c7e.png)

**AsciiDoc Plugin Version**: 2.8.3

**Kroki.io** is used for diagram rendering (use_kroki is enabled)

### Example for Reproduction

[asciidoc-issue-demo.zip](https://github.com/asciidoctor/asciidoctor-vscode/files/5136689/asciidoc-issue-demo.zip)

### Description

Hi,

as you can see in the example, a plantUML diagram included using the plantuml:: syntax does not render in the preview if it includes further plantuml files relative to itself. According to the docs this is syntactically correct. Also, exporting the example to HTML using the docToolchain works fine, so it seems to be good in general. The export is also in the zip file.

The preview of the plantUML plugin in VS Code can also display both diagrams.

However, the asciidoc preview looks as follows, reporting an error that it cannot find the 2nd level include:

![image](https://user-images.githubusercontent.com/61683053/91449558-8b112700-e87b-11ea-8705-e3cbe13ef6b3.png)

#### Observed behavior (1)

includes in plantuml files cause errors in the rendering process

#### Expected bevavior (1)

includes in plantuml files are processed correctly

I also want to report that there is another issue, likely in conjunction with the issue above:

#### Observed behavior (2)

When saving changes in a plantUML file while the asciidoc preview is open, the preview shows errors for all files until it's closed and re-opened

#### Expected behavior (2)

Saving changes in an included plantUML file does not break the preview

Kind regards
Jacques

anb0s:
I'm also using `!include` pattern and it works with own and public kroki-server. This was added in [asciidoc-kroki v0.9.1](https://github.com/Mogztter/asciidoctor-kroki/releases/tag/v0.9.1)

Please try the latest asciidoctor-vscode version...

anb0s:
I've just tried and think it should work with version [asciidoc-kroki v0.12.0](https://github.com/Mogztter/asciidoctor-kroki/releases/tag/v0.12.0) because of fixed [Resolve relative PlantUML includes from the diagram directory (not from the Asciidoctor base directory)](https://github.com/Mogztter/asciidoctor-kroki/issues/201) that should be integrated first wth PR #390 :)

jacques-ro:
[...]

Issue #203, opened 2019-05-26
Leaking file descriptors


jvalkeal:
After working with asciidoctor(using a preview) file I eventually get an error:

```
EMFILE: EMFILE: too many open files, open '/home/jvalkealahti/.vscode/extensions/joaompinto.asciidoctor-vscode-2.7.4/media/asciidoctor-editor.css'
```

Then I need to restart vscode for things to start working again. I changed system ulimit settings but it only delayed this error to popup.

Issue seems to be for this process:
```
jvalkea+ 26702 26649  9 07:45 tty2     00:00:05 /usr/share/code/code --nolazy --inspect=43713 /usr/share/code/resources/app/out/bootstrap-fork --type=extensionHost
```
When I open/close preview open file count keeps growing up and then at some it will hit a system limit.
```
jvalkealahti@cypher:~$ lsof -p 26702|wc
    557    5114   89673
jvalkealahti@cypher:~$ lsof -p 26702|wc
    697    6374  117522
jvalkealahti@cypher:~$ lsof -p 26702|wc
    837    7634  145371
```
Looking closely of `lsof` output it has same files opened multiple times, i.e:
```
code    26702 jvalkealahti  433r      REG               8,18     7952  2513326 /home/jvalkealahti/repos/jvalkeal/spring-statemachine/docs/src/reference/asciidoc/sm-repository.adoc
code    26702 jvalkealahti  434r      REG               8,18     9279  2490979 /home/jvalkealahti/repos/jvalkeal/spring-statemachine/docs/src/reference/asciidoc/samples/DocsJpaRepositorySampleTests1.java
code    26702 jvalkealahti  435r      REG               8,18     9279  2490979 /home/jvalkealahti/repos/jvalkeal/spring-statemachine/docs/src/reference/asciidoc/samples/DocsJpaRepositorySampleTests1.java
code    26702 jvalkealahti  436r      REG               8,18     9279  2490979 /home/jvalkealahti/repos/jvalkeal/spring-statemachine/docs/src/reference/asciidoc/samples/DocsJpaRepositorySampleTests1.java
code    26702 jvalkealahti  437r      REG               8,18     9279  2490979 /home/jvalkealahti/repos/jvalkeal/spring-statemachine/docs/src/reference/asciidoc/samples/DocsJpaRepositorySampleTests1.java
[...]

Issue #201, opened 2019-05-22
Scroll sync seems broken (was working in earlier versions)


rocketraman:
## Description

Using AsciiDoc plugin v2.7.3 with VS Code 1.34.0. Sometime recently, scroll syncing seems to be broken. I have verified the extension settings "Scroll Editor with Preview" and "Scroll Preview With Editor" are both checked.

## System Information

Fedora 29, VS Code 1.34.0

## To Reproduce

Steps to reproduce the issue:

1. Open an AsciiDoc file, and view the preview.
2. Click on some text in the source -- the preview does not scroll.
3. Click on some text in the preview -- the editor does not scroll.

## Additional Context

This used to work fine, but I don't know the version numbers when it worked.

rocketraman:
I see this is a duplicate of https://github.com/asciidoctor/asciidoctor-vscode/issues/182, closing. By the way, as mentioned in that issue, version 2.3.3 solves the problem on my machine as well.

MatteoCampinoti94:
Thanks for the report :)

Yeah we are aware that version 2.3.3 works correctly, I'm still investigating but I dont have a lot of free time atm so it might take a while. Sorry :(

Issue #126, opened 2018-11-13
Preview issue after update to 1.29


G3ph4z:
OS: Debian 9.6
vSCode: 1.29.0

I just updated my vscode from 1.28 to 1.29 and now, its broken.

When I try to open the preview I get this error message:
`command 'markdown.extension.togglePreview' not found`

joaompinto:
Could you please confirm that you are using the latest extension version which is 1.0.1 ?

That error seems to be related to a broken 1.0.0 update.

Thanks for reporting.

joaompinto:
Sorry, can you upgrade now to 1.0.2 ? Please make sure you have closed all code instances after updating the plugin. I am sure the "markdown.extension.togglePreview" was not present in 1.0.1, so a previous version of the extension is running somehow...

G3ph4z:
Ofc. I am on 1.0.2 right now, but it still doesn't work. I closed everything and reinstalled the plugin.

Version:
![image](https://user-images.githubusercontent.com/26697727/48473623-20534080-e7f9-11e8-82db-f1cc5d85b5d8.png)

Error message:
![image](https://user-images.githubusercontent.com/26697727/48473607-17fb0580-e7f9-11e8-860c-ae7d3f3dd9c6.png)

joaompinto:
Could you please provide the ouput from:
`grep -r markdown.extension.toggle ~/.vscode/extensions`

G3ph4z:
```
  grep -r markdown.extension.toggle ~/.vscode/extensions
/home/denes/.vscode/extensions/yzhang.markdown-all-in-one-1.7.0/package.json:                           "command": "markdown.extension.togglePreview",
/home/denes/.vscode/extensions/yzhang.markdown-all-in-one-1.7.0/package.json:                           "command": "markdown.extension.togglePreviewToSide",
/home/denes/.vscode/extensions/yzhang.markdown-all-in-one-1.7.0/out/src/preview.js:    context.subscriptions.push(vscode_1.commands.registerCommand('markdown.extension.togglePreview', () => {
/home/denes/.vscode/extensions/yzhang.markdown-all-in-one-1.7.0/out/src/preview.js:    }), vscode_1.commands.registerCommand('markdown.extension.togglePreviewToSide', () => {
```

BogdanI-FintechOS:
[...]

PR #390, opened 2021-03-03
updated asciidoc-kroki to 0.12.0

mariotoffia:
Hi again, I wonder if it is possible to release this - I understand that you don't want to make releases based on such as simple things that updating a reference.

However, there where a serious issue that @Mogztter has fixed in [asciidoctor-kroki](https://github.com/Mogztter/asciidoctor-kroki) that we use as excalidraw diagrams. Now he has enabled so we may render larger diagrams and we do no longer need to go through generate svg from diagram and update the repo. Instead we may render our diagrams inline in the asciidoc document.

Currently I've distributed a self compiled version of your excellent asciidoctor-vscode plugin!

Cheers,
 Mario :)

There are some other enhancement - see https://github.com/Mogztter/asciidoctor-kroki/releases/tag/v0.12.0

## Security

* Check Asciidoctor security level before reading or writing to disk. Please note that, preprocessing, kroki-fetch-diagram and kroki-plantuml-include are only available when safe mode is server or lower.

## Bug fixes

* Pass the diagram directory to the preprocessing functions (for clarity, rrename the parameter to diagramDir) - thanks @stenzengel
* Resolve relative PlantUML includes from the diagram directory (#201) - thanks @anb0s

## Improvements

* Make the max URI length configurable (default 4000) (#203)
* Execute the PlantUML preprocessor when type is c4plantuml (#126)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,7360 @@
 {
 	"name": "asciidoctor-vscode",
 	"version": "0.0.0",
-	"lockfileVersion": 1,
+	"lockfileVersion": 2,
 	"requires": true,
+	"packages": {
+		"": {
+			"version": "0.0.0",
+			"license": "MIT",
+			"dependencies": {
+				"@asciidoctor/core": "2.1.0",
+				"@asciidoctor/docbook-converter": "^2.0.0",
+				"@types/vscode": "^1.31.0",
+				"asciidoctor-kroki": "^0.12.0",
+				"file-url": "^3.0.0",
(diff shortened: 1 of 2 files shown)


anb0s:
That would be good, because it will fix the issue #361 👍

mariotoffia:
@danyill  or @joaompinto is there any chance that you do a release of this in a near future?

Cheers,
 Mario :)

joaompinto:
Will do a release this weekend. Thanks for the contribution.

PR #399, opened 2021-03-25
Fixed first-development missing JS

rongcuid:
Changed launch task dependency so it runs `webpack` prior to launching debug window. 

Previous behavior: `dist/index.js` and `dist/pre.js` were not compiled and included, and preview window in extension development host fail to load the scripts.

Current behavior: `webpack` is run before compiling, generating the required scripts.


rongcuid:
This is covered by #400

PR #407, opened 2021-05-09
adds xref autocomplete

s0lvang:
It looks for labels in all adoc files in the project and autocompletes when you write xref


joaompinto:
LGTM

PR #424, opened 2021-07-29
Added snippets for all document-level attributes

rben01:
Copied from here: https://docs.asciidoctor.org/asciidoc/latest/attributes/document-attributes-reference/


package-lock.json
```diff
@@ -1,7517 +1,8 @@
 {
 	"name": "asciidoctor-vscode",
-	"version": "0.0.0",
-	"lockfileVersion": 2,
+	"version": "2.9.0",
+	"lockfileVersion": 1,
```
Mogztter, line 4: Could you please revert this change? Alternatively, you can use npm >= 7 to regenerate the `package-lock.json` file (using version 2 format).

snippets/snippets.json
```diff
@@ -117,115 +107,795 @@
...
+			"",
+			"|${5:Cell 1, row 1}",
+			"|${6:Cell 2, row 1}",
+			"|${7:Cell 3, row 1}",
+			"|===",
+			"$0"
+		],
+		"description": "Insert table"
+	},
+	"Insert source highlighter": {
```
Mogztter, line 170: Maybe we should use the verb "Declare" or "Define" when we declare/define an attribute?

Mogztter:
This is really good, thanks!
I left a few comments, let me know what you think.

snippets/snippets.json
rben01: Yeah, there's probably something better than just "insert" everywhere. I think it would make sense to change the wording based on the type of the attribute, e.g., "Enable :hide-uri-scheme:" for a bool vs "choose :docinfo:" for an enum vs "set :author:" for an arbitrary string.

Mogztter: Yes that's a good idea but let's keep it for now, we can improve it later.

Mogztter:
And merged, thanks @rben01 🙌🏻

PR #441, opened 2021-09-12
WIP: Integration of asciidoctor-web-pdf instead of wkhtmltopdf

### LINKED ISSUE #433: Replace wkhtmltopdf with asciidoctor-web-pdf
See #239 for background and context. There are a pile of other issues which have been logged more recently I think that would also be resolved with this.

Although asciidoctor-web-pdf is still alpha, incorporating it now might provide valuable testing feedback and wkhtmltopdf is hardly a performant or even in a functional state for many use cases.

I've prototyped this and found it to work fine. The main downside is the need to download a fairly large binary and a slightly slower pdf output.

danyill:
Resolves #433.

Still working on this and need to think about acceptance criteria, testing on Mac/Windows, options which need to be added for asciidoctor-web-pdf etc.


rogierhofboer:
Please do not replace wkhtmltopdf (yet), but make asciidoctor-web-pdf an option next to it. I've tried asciidoctor-web-pdf and it had rendering errors with an included asciidoc generated from an OpenAPI definition and after those errors it did not render the rest of the document. the preview and wkhtmltopdf had no problems rendering the document.

danyill:
Thanks for the feedback. This is only a draft and perhaps we need to wait for further improvements in asciidoctor-web-pdf.

I'm also somewhat confident that the binary versions of asciidoctor-web-pdf also aren't working quite as I had hoped (there may be a memory leak, I need to examine further).

However I don't think retaining wkhtmltopdf is viable anymore. We are using an old binary that is not easily able to be updated and we have quite a few issues logged against it. 

There are many VS Code extensions which allow you to run an executable command against a file, for instance: 

* [Run On Save](https://marketplace.visualstudio.com/items?itemName=emeraldwalk.RunOnSave)
* [Terminal Commands](https://marketplace.visualstudio.com/items?itemName=adrianwilczynski.terminal-commands)
* [etc](https://marketplace.visualstudio.com/search?term=run%20command%20on%20file&target=VSCode&category=All%20categories&sortBy=Relevance)

Would these be reasonable alternatives for you?

Mogztter:
> However I don't think retaining wkhtmltopdf is viable anymore. We are using an old binary that is not easily able to be updated and we have quite a few issues logged against it.

Indeed, we rely on https://github.com/joaompinto/wkhtmltopdf/releases but I'm not sure how to update/publish a new release? @joaompinto if you have time, it would be great if you could tell us how to create a new release from the binaries https://wkhtmltopdf.org/downloads.html?

Alternatively, we could remove the download feature and tell users that they need to install the latest version from https://wkhtmltopdf.org/downloads.html?

> Please do not replace wkhtmltopdf (yet), but make asciidoctor-web-pdf an option next to it.

I agree, we should introduce this new feature as an option. I feel like users are somewhat happy with `wkhtmltopdf`. And, since Asciidoctor Web PDF is experimental we should give them the choice.

danyill:
Regrettably this PR is now so dated and I am unlikely to address it soon, so it would be better to start from scratch. I'll close on this basis.

PR #445, opened 2021-09-17
Run kroki extension within class, closes #444

### LINKED ISSUE #444: Asciidoc: Use_kroki setting change only effective after restarting VS code
## Description

Changes of the Use_kroki setting are only effective after restarting VS code.
## Environment
AsciiDoc v2.8.9

## To Reproduce

Create an asciidoc with an embedded diagram.
Check  Asciidoc: Use_asciidoctor_js 
Uncheck: Asciidoc: Use_kroki
Restart VS code
Preview the asciidoc (no diagrams should be rendered)
Check: Asciidoc: Use_kroki
Refresh the preview (diagrams should be rendered, but are not rendered)
Restart VS code
Refresh the preview (diagrams are rendered,)

danyill:
This means that changes to user preferences will immediately affect the next preview action.

That means that users can expect to still have to type some more text for a refresh which then includes kroki diagrams but they do not need to restart VS Code (which is such a confusing requirement that users may give up).

(It would be nicer to be able to have a hook to refresh the preview on a preference change but I haven't been able to see how to do that and that's probably a little wider than this issue.)

While I was there I renamed the file from `text-parser` to better represent the class `AsciidocParser`.
I think it makes sense for the registry and the processor to be available as object properties.

I will be happy to receive a review and make improvements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 # CHANGELOG
 
-## Next Version
+## Unreleased
 
-- Make `use_kroki` setting change effective without VS code restart (closes #444)
+- Make `useKroki` setting change effective without VS code restart (closes #444)
 
 ## 2.8.9
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- Make `useKroki` setting change effective without VS code restart (closes #444)
+- Make `use_kroki` setting change effective without VS code restart (closes #444)
 
 ## 2.8.9
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## Next Version
+
+- Make `use_kroki` setting change effective without VS code restart (closes #444)
+
 ## 2.8.9
 
 - Fix links in preview (closes #397)
--- a/src/asciidocParser.ts
+++ b/src/asciidocParser.ts
@@ -5,24 +5,21 @@ import { isNullOrUndefined } from 'util';
 
 const fileUrl = require('file-url');
 
-const asciidoctor = require('@asciidoctor/core')()
+const asciidoctor = require('@asciidoctor/core')
 const docbook = require('@asciidoctor/docbook-converter')
 const kroki = require('asciidoctor-kroki')
 const highlightjsAdapter = require('./highlightjs-adapter')
-highlightjsAdapter.register(asciidoctor)
-
-const use_kroki = vscode.workspace.getConfiguration('asciidoc', null).get('use_kroki');
-if (use_kroki) {
-  kroki.register(asciidoctor.Extensions);
-}
 
 export class AsciidocParser {
     public html: string = '';
     public document = null;
+    public adProcessor = null
+    public registry = null
     private ext_path = vscode.extensions.getExtension('asciidoctor.asciidoctor-vscode').extensionPath;
     private stylesdir = path.join(this.ext_path, 'media')
 
     constructor(private readonly filename: string, private errorCollection: vscode.DiagnosticCollection = null) {
+      this.adProcessor = asciidoctor()
     }
 
     public getAttribute(name: string) {
@@ -53,8 +50,18 @@ export class AsciidocParser {
           this.errorCollection.clear();
         }
 
-        const memoryLogger = asciidoctor.MemoryLogger.create();
-        asciidoctor.LoggerManager.setLogger(memoryLogger);
+        const memoryLogger = this.adProcessor.MemoryLogger.create()
+        this.adProcessor.LoggerManager.setLogger(memoryLogger)
+
+        this.registry = this.adProcessor.Extensions.create()
+
+        highlightjsAdapter.register(this.registry)
+
+        const use_kroki = vscode.workspace.getConfiguration('asciidoc', null).get('use_kroki')
+
+        if (use_kroki) {
+          kroki.register(this.registry)
+        }
 
         var attributes = {};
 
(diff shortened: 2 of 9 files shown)


Mogztter:
> (It would be nicer to be able to have a hook to refresh the preview on a preference change but I haven't been able to see how to do that and that's probably a little wider than this issue.)

💯 
We should create an issue to track progess. We could even refresh the preview so that users don't have to type some more test to trigger a refresh.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## Next Version
```
Mogztter, line 3: suggestion: Use "Unreleased" instead? I believe "Unreleased" is more common than "Next Version" but to be honest I'm fine with both 😄

src/asciidocParser.ts
```diff
@@ -53,8 +50,18 @@ export class AsciidocParser {
...
-        const memoryLogger = asciidoctor.MemoryLogger.create();
-        asciidoctor.LoggerManager.setLogger(memoryLogger);
+        const memoryLogger = this.adProcessor.MemoryLogger.create()
+        this.adProcessor.LoggerManager.setLogger(memoryLogger)
+
+        this.registry = this.adProcessor.Extensions.create()
+
+        highlightjsAdapter.register(this.registry)
+
+        const use_kroki = vscode.workspace.getConfiguration('asciidoc', null).get('use_kroki')
```
Mogztter, line 60: nitpick: use camelCase

```suggestion
        const useKroki = vscode.workspace.getConfiguration('asciidoc', null).get('use_kroki')
```

(it was already in snake_case butit should be camelCase)

src/asciidocParser.ts
```diff
@@ -5,24 +5,21 @@ import { isNullOrUndefined } from 'util';
...
-
-const use_kroki = vscode.workspace.getConfiguration('asciidoc', null).get('use_kroki');
-if (use_kroki) {
-  kroki.register(asciidoctor.Extensions);
-}
 
 export class AsciidocParser {
     public html: string = '';
     public document = null;
+    public adProcessor = null
```
Mogztter, line 16: (nitpick): since we are in an `AsciidocParser` we could name it `processor`? as an alternative, I would use the complete name `asciidocProcessor`. `ad` can be confusing if you are not familiar with the code base ("is that a typo, it should be addProcessor no?" 😆)

danyill: :+1:

src/asciidocParser.ts
danyill: :+1:

CHANGELOG.md
danyill: Sure.

src/asciidocParser.ts
```diff
@@ -53,8 +50,18 @@ export class AsciidocParser {
...
         }
 
-        const memoryLogger = asciidoctor.MemoryLogger.create();
-        asciidoctor.LoggerManager.setLogger(memoryLogger);
+        const memoryLogger = this.processor.MemoryLogger.create()
+        this.processor.LoggerManager.setLogger(memoryLogger)
+
+        this.registry = this.processor.Extensions.create()
+
+        // highlightjsAdapter.register(this.processor)
```
danyill, line 58: I disabled this because we're still thinking through it and also because it wasn't working with the registry.

danyill:
Thank you for your review. I am grateful for all comments and happy to receive nitpicks. We will converge on a way of doing things quickly.

I have now introduced new inconsistencies by camelCasing `useKroki` consistently everywhere :wink: 

But we can tidy them up in time.

src/asciidocParser.ts
Mogztter: Sorry I was only talking about the variable name in the JavaScript code. We probably shouldn't rename the name of the configuration option.
If we do that I guess we will need to "migrate" existing configuration no?

src/asciidocParser.ts
Mogztter: > and also because it wasn't working with the registry.

The reason is that the highlight.js adapter is not an Asciidoctor extension but a custom syntax highlighter. So we need to pass the "asciidoctor instance" to be able to call `asciidoctor.SyntaxHighlighter.for('highlight.js')`.

The syntax highlighter registry and the extension registry are two things

src/asciidocParser.ts
danyill: Yes, I guess it's unnecessary at this time. The settings are a mix of different casing styles. I will revert.

src/asciidocParser.ts
```diff
@@ -53,8 +50,18 @@ export class AsciidocParser {
...
         }
 
-        const memoryLogger = asciidoctor.MemoryLogger.create();
-        asciidoctor.LoggerManager.setLogger(memoryLogger);
+        const memoryLogger = this.processor.MemoryLogger.create()
+        this.processor.LoggerManager.setLogger(memoryLogger)
+
+        this.registry = this.processor.Extensions.create()
+
+        this.processor.SyntaxHighlighter.for('highlight.js')
```
Mogztter, line 58: I think that the highlight.js syntax highlighter "adapter" should only be registered once.
Furthermore, the correct way to register it is:

```js
highlightjsAdapter.register(this.processor)
```

So I would move the above statement in the `constructor`:

```js
constructor(private readonly filename: string, private errorCollection: vscode.DiagnosticCollection = null) {
  this.processor = asciidoctor()
  highlightjsAdapter.register(this.processor)
}
```

src/asciidocParser.ts
```diff
@@ -5,24 +5,21 @@ import { isNullOrUndefined } from 'util';
...
-const use_kroki = vscode.workspace.getConfiguration('asciidoc', null).get('use_kroki');
-if (use_kroki) {
-  kroki.register(asciidoctor.Extensions);
-}
 
 export class AsciidocParser {
     public html: string = '';
     public document = null;
+    public processor = null
+    public registry = null
```
Mogztter, line 17: I don't think we need to declare the `registry` as a class attribute.
It should be enough to declare a local variable in the `convert_using_javascript` function no?

src/asciidocParser.ts
danyill: Thank you for your patience. Done.

Mogztter (approved):
Looks good to me 👍🏻

danyill:
Thank you for the review @Mogztter.

Because of the substantial linting changes, I'm going to close this PR and open a new PR with this rebased on top of the linting changes. I hope that's appropriate. It's a little unclear what house rules should be for this and I am inexperienced in this.

Mogztter:
You can rebase and force push your branch. You don't need to create a new pull request.

As long as you didn't merge changes from other branches you should be fine with a rebase.

danyill:
> You can rebase and force push your branch. You don't need to create a new pull request.
> 
> As long as you didn't merge changes from other branches you should be fine with a rebase.

I don't think that's possible, I can't reopen it and the tooltip text suggests it's because of the force push. Sorry.

image

Mogztter:
No worries, you can keep this tip in mind for next time.

Issue #435, opened 2021-08-24
Clicking links to local files inside a preview window does not work


lkydev:
## Description

Clicking links to local files inside a preview window does not work.

## System Information

Version: 1.59.1 (system setup)
Commit: 3866c3553be8b268c8a7f8c0482c0c0177aa8bfa
Date: 2021-08-19T11:56:46.957Z
Electron: 13.1.7
Chrome: 91.0.4472.124
Node.js: 14.16.0
V8: 9.1.269.36-electron.0
OS: Windows_NT x64 10.0.17763

Running asciidoctor extension 2.8.9

## To Reproduce

1. Create a folder C:\Users\\`<N>`\Downloads\test where `<N>` is your user account name. (In other words, create a folder named 'test' inside your Download directory.)
2. Open the created folder with VSCode. When asked whether you trusted the authors, click `Yes, I trust the authors`.
3. Create an asciidoc file C:\Users\\`<N>`\Downloads\test\test1.adoc. (Remember to substitute your user account name for `<N>`, including the content below): 

```
= This is test1.adoc

link:test2.adoc[1]
link:.\test2.adoc[2]
link:C:\Users\<N>\Downloads\test\test2.adoc[3]
link:file:///C:\Users\<N>\Downloads\test\test2.adoc[4]
```

4. Create another asciidoc file C:\Users\\`<N>`\Downloads\test\test2.adoc:
```
= This is test2.adoc
```

5. Open the preview window for test1.adoc. Click any links. The first two unexpectedly open a web browser window (see the screenshot). The last two do nothing.

## Screenshots & Files

This is what the browser window looks like

![image](https://user-images.githubusercontent.com/22630134/130560330-573f15d6-6be6-427c-82c1-846e83dc6178.png)

## Additional Context
I expect a tab to be opened for test2.adoc when any link inside test1.adoc is clicked in the preview window.

danyill:
The fix in #454 would be functional but doesn't provide strong integration with the VS Code security model.

I think previously these links were rewritten with `vscode-resource` (see [here](https://github.com/asciidoctor/asciidoctor-vscode/issues/397#issuecomment-806984770)) but some time in the last 6 months this became non-functional (perhaps earlier, perhaps on the insiders built? See [here](https://github.com/microsoft/vscode/issues/102959#issuecomment-663340395))

[...]

PR #400, opened 2021-03-25
Fix preview opening relative link

rongcuid:
Fixes #397. I haven't tested very comprehensively, but I tried a few common cases (that I encounter).

Issue #397, opened 2021-03-22
Unable to open relative links


rongcuid:
## Description

In preview window, when I click on a link to relative file, it does not open (either InEditor or InPreview).

## System Information

```
Version: 1.54.3
Commit: 2b9aebd5354a3629c3aba0a5f5df49f43d6689f8
Date: 2021-03-15T10:55:24.277Z
Electron: 11.3.0
Chrome: 87.0.4280.141
Node.js: 12.18.3
V8: 8.7.220.31-electron.0
OS: Linux x64 5.8.0-7642-generic
```

## To Reproduce
1. In a directory, create `A.adoc` and `B.adoc`. Fill in random contents.
2. In `A.adoc`, create a relative link by:
    * `link:B.adoc`, or
    * `<<B.adoc#>>`
3. Try to click on link in preview window

Any other combination of `outfilesuffix` or `relfilesuffix` result in the same behavior.

rongcuid:
OK, I triangulated the problem, but I am unable to solve it due to unfamiliarity with VSCode or TS:

`src/features/preview.ts:133`, here the extension registers a WebView callback `onDidReceiveMessage` which handles the link clicking and various other features. However, this closure is never executed. I don't know if it failed to register or something.

rongcuid:
OK, it doesn't make sense. The callback seems to be fine, but nothing in `index.ts` gets executed. There lies the code to actually send the `clickLink` events, etc.

rongcuid:
I found the issue:
`index.ts:137` supports only `file:` and `vscode-resource:` protocol. However, using the Webview debug window, I find that the links are actually `vscode-webview-resource:` protocol, and seem to follow a different format. 

I checked the VSCode Markdown extension, and find that instead of processing _inside_ the `index.ts` script which is run in Webview, it instead lets the backend resolve the URI. I am working on refactoring this part to do the same. This modification will also allow passthrough, though I am not sure what it may be used for.

For reference:
https://github.com/microsoft/vscode/blob/main/extensions/markdown-language-features/preview-src/index.ts

PR #448, opened 2021-09-17
Enforce code style via linter (part of #446)

### LINKED ISSUE #446: Enforce a code style (via a linter) to make the code consistent
We could use [ESLint](https://github.com/eslint/eslint) with https://github.com/typescript-eslint/typescript-eslint or [Standard](https://standardjs.com/) I don't really mind but I think it's important to enforce a code style.

danyill:
This will eventually close #446 but is draft for now.

Updating dependencies to use `standard` for lining has resulted in `package-lock` being rewritten to a new version.

Immediate output is: `3242 problems`.
After auto-fixing we get `1269 problems`.
Most seem to be spaces and tabs and things which are never used.
I'm happy to go through these.

I guess ideally we would wait until we had no open PRs before doing something like this but this PR just allows us to understand the implications.

WDYT?


danyill:
I had thought it would be good to set up a test framework prior to making "big changes" but the extension does still seem to work at this point. I have no experience as to whether linters are likely to break working code.

Mogztter:
> Updating dependencies to use standard for lining has resulted in package-lock being rewritten to a new version.

The reason `package-lock.json` contains a lot of changes is because you are using npm <= 6. You should use npm 7 on this project (with Node >= 14). 
Currently, the lock file is using version 2 (you can see it in the diff: `lockfileVersion: 2`).
If you install dependencies with npm <= 6 it will regenerate the lock file using version 1.

> I guess ideally we would wait until we had no open PRs before doing something like this

Yes, otherwise we will need to rebase and resolve conflicts.

> I had thought it would be good to set up a test framework prior to making "big changes" but the extension does still seem to work at this point.

Indeed that would be good but it will take time until we have a good test overage.

>  I have no experience as to whether linters are likely to break working code.

I believe that `standard --fix` is safe unless there's a bug in `standard`.
Having said that, it's probably a good idea to perform extensive (manual) tests once all lint issues are fixed but before merging this pull request.

danyill:
Something is happening -- and it certainly took some time.

```
23:03 $ npm run lint

> asciidoctor-vscode@0.0.0 lint
> eslint 'src/**/*.ts' 'src-preview/**/*.ts' --format unix

/home/mulhollandd/Documents/asciidoctor-vscode/src/asciidocExtensions.ts:12:31: 'Thenable' is not defined. [Error/no-undef]
/home/mulhollandd/Documents/asciidoctor-vscode/src/asciidocExtensions.ts:21:30: 'Thenable' is not defined. [Error/no-undef]
/home/mulhollandd/Documents/asciidoctor-vscode/src/asciidocExtensions.ts:51:36: 'Thenable' is not defined. [Error/no-undef]
/home/mulhollandd/Documents/asciidoctor-vscode/src/commands/openDocumentLink.ts:40:30: Expected 'undefined' and instead saw 'void'. [Error/no-void]
/home/mulhollandd/Documents/asciidoctor-vscode/src/commands/openDocumentLink.ts:42:21: Expected 'undefined' and instead saw 'void'. [Error/no-void]
/home/mulhollandd/Documents/asciidoctor-vscode/src/features/workspaceSymbolProvider.ts:13:30: 'Thenable' is not defined. [Error/no-undef]
/home/mulhollandd/Documents/asciidoctor-vscode/src/features/workspaceSymbolProvider.ts:99:47: 'Thenable' is not defined. [Error/no-undef]
/home/mulhollandd/Documents/asciidoctor-vscode/src/features/workspaceSymbolProvider.ts:135:59: 'Thenable' is not defined. [Error/no-undef]
/home/mulhollandd/Documents/asciidoctor-vscode/src/security.ts:23:91: 'Thenable' is not defined. [Error/no-undef]
/home/mulhollandd/Documents/asciidoctor-vscode/src/security.ts:29:57: 'Thenable' is not defined. [Error/no-undef]
/home/mulhollandd/Documents/asciidoctor-vscode/src/security.ts:59:99: 'Thenable' is not defined. [Error/no-undef]
/home/mulhollandd/Documents/asciidoctor-vscode/src/security.ts:72:63: 'Thenable' is not defined. [Error/no-undef]
```

The `Thenable` didn't have a type definition. I took one from the Markdown repo and placed in `src/typings`. But I can't seem to get eslint to recognise it. And I'm a little uncertain about the undefined/void at the moment.

I haven't done any significant testing yet and I would want to spend some time on it because there were quite a few non-automatic changes and I may have introduced errors (particularly around `isNullOrUndefined` where I decided it shouldn't always need to be written as the or of both expressions but I may not have always chosen correctly).

Mogztter:
> The Thenable didn't have a type definition. I took one from the Markdown repo and placed in src/typings. But I can't seem to get eslint to recognise it.

I think we need to add:

```json
{
  "env": {
    "node": true,
    "es6": true
  }
}
```

in the `.eslintrc.js` file

> And I'm a little uncertain about the undefined/void at the moment.

I can't understand why we do such things:

```js
Promise.resolve(void 0)
        .then(() => vscode.commands.executeCommand('vscode.open', resource))
        .then(() => void 0)
```

`vscode.commands.executeCommand('vscode.open', resource)` is asynchronous and `void 0` is a function that returns `undefined` but we could write:

```js
public async execute(args: OpenDocumentLinkArgs) {
  // ...
  await vscode.commands.executeCommand('vscode.open', resource)
  return
}
```

That's almost what the Markdown plugin is doing: https://github.com/microsoft/vscode/blob/2f33470e28c1a3440b2c5e0ea9205b1d0a1e6f1b/extensions/markdown-language-features/src/commands/openDocumentLink.ts#L66-L82

Mogztter:
@danyill I've replaced `Thenable` by `Promise` (included in Node typing) and replaced `Promise.resolve(void 0)` by async/await.

Mogztter:
I've also enabled GitHub Action CI workflow (lint and compile) on pull requests

danyill:
Thank you, that's marvellous. I'd like to spend an hour or two on the weekend seeing if I can find obvious flaws and then hopefully we can merge a few things.

danyill:
I did some tests.
I only tested on Linux.
I only tested on VS Code 1.60.1.

* [x] Check that syntax highlighting still appears
* [x] Check that preview runs (locked and preview to the side)
* [x] Check for no errors in web view (Developer Tools) for a simple document
* [x] Check that document symbols exist
* [x] Check that I can create a simple Export to PDF with wkhtmltopdf (*Initial Fail* had to expand null/undefined)
* [x] Check that I can create a simple Export to PDF with asciidoctor-pdf (*Initial Fail*, looks like the option check wasn't appropriately case sensitive)
* [x] Check that I can save to HTML
* [x] Check that I can save to Docbook
* [x] Check that Docbook 4.5 conversion works
* [x] Check that with security disabled I see admonitions
* [x] Check that error diagnostics work by adding an incorrect heading level
* [x] Check that scroll sync works and can be disabled and re-enabled with options (*Initial Fail*, it seems this option requires an extension restart but they do work. I'll log an issue for this soon)
* [ ] Check that links work correctly (*_FAIL_* they don't but I think there are issues logged about this so I haven't fixed here)
* [x] Check that using Asciidoctor standard styling works (*Initial Fail*, requires preview to be closed and then reopened)
* [x] Check that Kroki works
* [x] Check that an image paste works
* [x] Test that showing or hiding front matter setting works
* [x] Check that Asciimath works in the preview
* [x] Check that snippet insertion works
* [x] Check that the outline is populated and I can go to sections with Ctrl+Shift+O

In general things seem to be working OK, somewhat obscured by the number of bugs and usability issues on various features.
I think any breakages will not be difficult to fix if they occur.

If you're happy with this PR @Mogztter feel free to merge away otherwise I'll be happy to look into any aspect in more detail.

Mogztter:
>  Check that links work correctly (FAIL they don't but I think there are issues logged about this so I haven't fixed here)

From what I seen in the code (before your changes) it won't work because the regular expressions are designed to work with Markdown syntax (not AsciiDoc syntax).

https://github.com/asciidoctor/asciidoctor-vscode/blob/74bce1d292f36a1672d6df608d1aa4d1e8a85720/src/features/documentLinkProvider.ts#L53-L55

danyill:
Ah yes, those too.
I was thinking of links in the WebView which I think used to work...

danyill:
> If you can confirm what exactly is not working with links (to make sure that this is related to existing issues) that would be great. Then, I think we can safely merge this pull request.

I've had a look, I think it's unrelated. 

Links within the preview were only recently added (see #397 and #400) and worked fine previously. I think the Webview is now handled differently due to upstream changes and file links end up with a prefix of `https://file+.vscode-resource.vscode-webview.net/`. This doesn't work with the previously link handling strategy so we need to make sure these are handled differently so they are resolved correctly.

https://github.com/asciidoctor/asciidoctor-vscode/blob/74bce1d292f36a1672d6df608d1aa4d1e8a85720/src-preview/index.ts#L143-L155

At least half a fix looks something like the following where we make sure we don't pass through file links and then ensure that it is handled by VS code to resolve.

```js
			let hrefText = node.getAttribute('data-href');
			if (!hrefText) {
				// Pass through known schemes
				if (passThroughLinkSchemes.some((scheme) => node.href.startsWith(scheme) && !node.href.startsWith('https://file+.vscode-resource.vscode-webview.net/'))) {
					return;
				}
				hrefText = node.getAttribute('href');
			}

			// If original link doesn't look like a url, delegate back to VS Code to resolve
			if (!/^[a-z\-]+:/i.test(hrefText) || hrefText.startsWith('http://https//file+.vscode-resource.vscode-webview.net')) {
				messaging.postMessage('clickLink', { href: hrefText });
				event.preventDefault();
				event.stopPropagation();
				return;
			}
```



I'll look into this as part of #435.

Mogztter:
Perfect, then we are good to go 😀

PR #451, opened 2021-09-19
Load Highlight.js from the extension

Mogztter:
## What I did

- Add Highlight.js 11.2.0 in `media/highlightjs`
- Load Highlight.js from the extension when using the preview (i.e., when using the WebView)
- Fallback to the built-in Highlight.js highlighter when the WebView is not available (for instance, when we use the export feature)
- Update the CSP to allow the inline script that highlight every elements (`'sha256-ZrDBcrmObbqhVV/Mag2fT/y08UJGejdW7UWyEsi4DXw='`)
- Update the CSP to allow `vscode-resource:`
- Add a task to copy Highlights.js assets from `node_modules` to `media/highlightjs`

Instead of allowing `vcode-resource:`, we could also add the variable `${webview.cspSource}`. When testing locally, the value of this variable is `https://*.vscode-webview.net`.

Not sure if it's more secure or not... it does not look like we can define a CSP to only allow resources that are loaded from our extension. Having said that, I think it's pretty safe because the user would have to install a VS Code extension that contains malicious code. And if so, then I guess it's probably easier to just execute code than to add a script in the webview of the AsciiDoc extension 😅


danyill:
I shall be happy to take a look, learn something and provide a few comments. Thanks!

Mogztter:
@danyill to make it easier to review, I can probably split my commit into:

1. add Highlight.js (copy the whole Highlight.js distribution in the `media/` directory)
2. use/integrate the local copy in the preview

Alternatively, we could ignore `media/highlightjs` (using `.gitignore`) and copy files just before the packaging.

danyill:
I don't mind reviewing it as is :wink:

src/highlightjs-adapter.ts
```diff
@@ -1,36 +1,36 @@
...
+    const highlightjsScriptResource = vscode.Uri.file(path.join(context.extensionPath, 'media', 'highlightjs', 'highlight.min.js'))
+    return `<script src="${webviewPanel.webview.asWebviewUri(highlightjsScriptResource)}"></script>
 ${languageScripts}
 <script>
 if (!hljs.initHighlighting.called) {
   hljs.initHighlighting.called = true
-  ;[].slice.call(document.querySelectorAll("pre.highlight > code")).forEach(function (el) { hljs.highlightBlock(el) })
+  ;[].slice.call(document.querySelectorAll("pre.highlight > code")).forEach(function (el) { hljs.highlightElement(el) })
 }
 </script>`
```
danyill, line 34: Will this ever be called if security is not set to disabled?

danyill:
> Alternatively, we could ignore media/highlightjs (using .gitignore) and copy files just before the packaging.

I think this would be good to avoid repository bloat, it seems a little unnecessary to commit the files to the repo if we never really intend to modify them.

danyill (changes requested):
This is quite delightful. Thank you for this contribution, I have learnt something from studying it and am rather humbled to be invited to review. :blush: 

I worry a little about #448 and whether this might make it quite hard to merge.

I think the one thing we must do in this PR is include the bundling of the new highlight.js resources into the `vscode:prepublish` or `package` npm tasks or they won't be included in the release via CI. I would probably put them in the `vscode:prepublish` task because that's where the other tasks are done and then a `vsce package` will still work I think.

src/features/previewContentProvider.ts
```diff
@@ -185,17 +186,17 @@ export class AsciidocContentProvider {
   private getCspForResource(resource: vscode.Uri, nonce: string): string {
     switch (this.cspArbiter.getSecurityLevelForResource(resource)) {
     case AsciidocPreviewSecurityLevel.AllowInsecureContent:
-      return `<meta http-equiv="Content-Security-Policy" content="default-src 'none'; img-src vscode-resource: http: https: data:; media-src vscode-resource: http: https: data:; script-src 'nonce-${nonce}'; style-src vscode-resource: 'unsafe-inline' http: https: data:; font-src vscode-resource: http: https: data:;">`;
+      return `<meta http-equiv="Content-Security-Policy" content="default-src 'none'; img-src vscode-resource: http: https: data:; media-src vscode-resource: http: https: data:; script-src vscode-resource: 'nonce-${nonce}' 'sha256-ZrDBcrmObbqhVV/Mag2fT/y08UJGejdW7UWyEsi4DXw='; style-src vscode-resource: 'unsafe-inline' http: https: data:; font-src vscode-resource: http: https: data:;">`;
```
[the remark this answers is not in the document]
Mogztter: You're absolutely correct.
I didn't want to introduce too many changes but this part should be completly rewritten.

As a first step, I can probably declare `'sha256-ZrDBcrmObbqhVV/Mag2fT/y08UJGejdW7UWyEsi4DXw='` as a constant with a descriptive name.

src/highlightjs-adapter.ts
Mogztter: What do you mean by "if security is not set to disabled?"

The hash `'sha256-ZrDBcrmObbqhVV/Mag2fT/y08UJGejdW7UWyEsi4DXw='` is declared in all CSP (even when the security level is "strict").

package.json
```diff
@@ -453,6 +453,7 @@
 		]
 	},
 	"scripts": {
+		"copy:assets": "cp node_modules/@highlightjs/cdn-assets/highlight.min.js media/highlightjs && cp -r node_modules/@highlightjs/cdn-assets/languages media/highlightjs && cp -r node_modules/@highlightjs/cdn-assets/styles media/highlightjs",
```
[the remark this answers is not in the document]
Mogztter: Is there a development/debug mode? Currently, I'm doing, `npm run package && code --install-extension *.vsix && code .` to test my changes.

danyill:
Oh that's too difficult!

If you open the extension folder in VS code and you go Run/Debug it will
open the extension host running the extension in another window and you can
put breakpoints in the first VS code editor instance.

Mogztter:
> Oh that's too difficult!
If you open the extension folder in VS code and you go Run/Debug it will open the extension host running the extension in another window and you can put breakpoints in the first VS code editor instance.

We should definitely add that in the contributing guide or add a section in the README to explain how to troubleshoot/debug the extension.

I was asking because then we need to run the task `npm run copy:assets` before doing that (otherwise the syntax highlighting with Highlight.js won't work).

package.json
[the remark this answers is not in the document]
Mogztter: I've added the `copy:assets` in a `prepackage` task.

package.json
```diff
@@ -2,7 +2,7 @@
 	"name": "asciidoctor-vscode",
 	"displayName": "%displayName%",
 	"description": "%description%",
-	"version": "0.0.0",
+	"version": "2.9.0-dev",
```
Mogztter, line 5: it makes things easier when you want to install the extension locally (otherwise VS Code will complain that version 0.0.0 cannot be installed)

Mogztter:
> I worry a little about #448 and whether this might make it quite hard to merge.

I don't mind, we can merge #448 first and I will rebase my work or the other way around.

Mogztter:
@danyill ready for round two 😉

src/highlightjs-adapter.ts
```diff
@@ -1,36 +1,36 @@
-// from https://github.com/cdnjs/SRIs/blob/master/highlight.js/9.18.3.json
-import * as sri from './highlightjs-sri-9.18.3.json'
+import * as vscode from 'vscode'
+import * as path from 'path'
+import { WebviewPanel } from "vscode";
```
danyill, line 3: We don't need this, right?

```suggestion
```

danyill:
> > Oh that's too difficult!
> > If you open the extension folder in VS code and you go Run/Debug it will open the extension host running the extension in another window and you can put breakpoints in the first VS code editor instance.
> 
> We should definitely add that in the contributing guide or add a section in the README to explain how to troubleshoot/debug the extension.
> 
> I was asking because then we need to run the task `npm run copy:assets` before doing that (otherwise the syntax highlighting with Highlight.js won't work).

Yes, I have provided a comment on what I _think_ is required.

src/highlightjs-adapter.ts
danyill: I was having a moment, all good!

danyill:
This is looking good to me.
I'm sorry if I'm a little slow, during the week there are sometimes many tasks.
I hope to look at the code style PR this weekend - I want to spend an hour or two trying to finding breakages.

danyill (approved):

package.json
```diff
@@ -453,12 +453,14 @@
...
 	},
 	"scripts": {
+		"copy:assets": "cp node_modules/@highlightjs/cdn-assets/highlight.min.js media/highlightjs && cp -r node_modules/@highlightjs/cdn-assets/languages media/highlightjs && cp -r node_modules/@highlightjs/cdn-assets/styles media/highlightjs",
 		"vscode:prepublish": "npm run compile && npm run webpack",
 		"webpack": "npm run webpack-preview",
 		"webpack-preview": "webpack --mode production --config webpack.config.preview.js",
 		"webpack-preview-dev": "webpack --config webpack.config.preview.js --mode development",
 		"compile": "tsc -p ./",
 		"package": "vsce package",
+		"prepackage": "npm run copy:assets",
```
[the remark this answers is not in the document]
danyill: Also if we include it in `compile` then it should function correctly in CI.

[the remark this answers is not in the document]
Mogztter: My understanding is that `asciidoctor-vscode/.vscode/launch.json` is only used in local development.
We definitely want to run `npm run copy:assets` but I don't think we should run this task when we want to compile the code.

The task `prepackage` will run before the task `package` automatically. That's a built-in feature of npm scripts, see: https://docs.npmjs.com/cli/v7/using-npm/scripts#pre--post-scripts

Basically, when I run `npm run package` it will run the `prepackage` task first.

> Also if we include it in compile then it should function correctly in CI.

When we release a new version, we do:

https://github.com/asciidoctor/asciidoctor-vscode/blob/74bce1d292f36a1672d6df608d1aa4d1e8a85720/.github/workflows/release.yml#L40-L41

So, assets will be copied before creating the extension bundle.

> (Note: It seems like the workflow is not quite right at the moment because we currently have to manually call npm run webpack or npm run webpack-dev. Ideally a developer starting out wouldn't need to know this but we should at least not make it worse here I guess.)

Could you expand on that? We need to run `npm run webpack` after (or before?) `npm run compile`?

> ... or npm run webpack-dev

You meant `npm run webpack-preview-dev` ?
If the webpack task is only used in development I would probably always add the `--mode development` flag. Do you know if there are drawbacks? Maybe it's a bit slower to use the development mode?


To be honest, I think it would be better to consolidate the scripts and use a `npm run dev` or `npm start` to run/use the extension in local development.

Also, I don't think git commands/aliases belong to the `scripts`, so I would also remove `"prune-branches": "git remote prune origin"` and `"clean": "git clean -fdX"`.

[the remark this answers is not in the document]
Mogztter: There were too many things to fix, I've created a new pull request: https://github.com/asciidoctor/asciidoctor-vscode/pull/452

src/highlightjs-adapter.ts
Mogztter: You are correct, I'm using `vscode.WebviewPanel` so `import * as vscode from 'vscode'` is sufficient.

package.json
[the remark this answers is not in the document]
danyill: > The task prepackage will run before the task package automatically. That's a built-in feature of npm scripts, see: https://docs.npmjs.com/cli/v7/using-npm/scripts#pre--post-scripts

Thanks!

> There were too many things to fix, I've created a new pull request: #452

You are on :fire:, this is the most :heart: this extension has seen in many days. :+1: 

I locally merged this #452 on top of this PR to see if it all came together.

With these changes, the extension host doesn't run for me when I hit run/debug on the extension. I think this is because `npm watch` doesn't exit and the extension host expects it to (called via the prelaunch task in `launch.json`).

image

>Also, I don't think git commands/aliases belong to the scripts, so I would also remove "prune-branches": "git remote prune origin" and "clean": "git clean -fdX".

I think we should retain a clean task that removes build artefacts (in `media/highlightjs` and `dist`) but I agree that git commands are the wrong idea.

> You meant npm run webpack-preview-dev ?
> If the webpack task is only used in development I would probably always add the --mode development flag. Do you know if there are drawbacks? Maybe it's a bit slower to use the development mode?

The webpack commands compile the js which is run within the webview preview so the files generated in `dist` need to produced for either locally debugging/developing the extension and for packaging. If the `media/highlightjs` and `dist` files aren't present when the extension runs there will be errors in the webview because they don't exist.

When #452 is merged, I think we need to include `copy:assets` into the `dev` command and then I think it all works nicely. Both the `dist` folder and the `highlightjs` resources are then available for running locally:

```js
    "scripts": {
        "copy:assets": "cp node_modules/@highlightjs/cdn-assets/highlight.min.js media/highlightjs && cp -r node_modules/@highlightjs/cdn-assets/languages media/highlightjs && cp -r node_modules/@highlightjs/cdn-assets/styles media/highlightjs",
        "dev": "webpack --config webpack.config.preview.js --mode development && npm run copy:assets",
```

And then with the "prepackage" command everything should be created correctly for releases if I (now) understand correctly.

Mogztter:
@danyill Rebased on master but https://github.com/asciidoctor/asciidoctor-vscode/pull/460 should go first

package.json
```diff
@@ -453,16 +453,20 @@
...
 	},
 	"scripts": {
-		"dev": "webpack --config webpack.config.preview.js --mode development && npm run compile",
-		"compile": "tsc -p ./",
-		"package": "webpack --config webpack.config.preview.js --mode production && npm run compile && vsce package",
+		"copy-assets": "cp node_modules/@highlightjs/cdn-assets/highlight.min.js media/highlightjs && cp -r node_modules/@highlightjs/cdn-assets/languages media/highlightjs && cp -r node_modules/@highlightjs/cdn-assets/styles media/highlightjs",
+		"dev": "npm run build",
+		"build": "npm run copy-assets && npm run build-ext && npm run build-preview",
+		"build-preview": "webpack --config webpack.config.preview.js --mode production",
+		"build-ext": "tsc -p ./",
```
Mogztter, line 460: I took inspiration from https://github.com/microsoft/vscode/blob/a4342f3addc29f405a5cb4aab359b0b98223411e/extensions/markdown-language-features/package.json#L344-L352 and renamed the tasks to "build" (instead of "compile").

As a result, I included the "copy assets" task in the "build" task.

We are now using `--mode production` even when running the dev or test task. That's what the Markdown extension is doing. It simplify things and, as far as I know, it has no impact since we can still debug the extension using VS code... right?

Mogztter, line 460: I took inspiration from https://github.com/microsoft/vscode/blob/a4342f3addc29f405a5cb4aab359b0b98223411e/extensions/markdown-language-features/package.json#L344-L352 and renamed the tasks to "build" (instead of "compile").

As a result, I included the "copy assets" task in the "build" task.

We are now using `--mode production` even when running the dev or test task. That's what the Markdown extension is doing. It simplify things and, as far as I know, it has no impact since we can still debug the extension using VS code... right?

danyill: Agree

stenzengel:
@Mogztter Thanks for this contribution. Is there the possibility to use a custom theme in the preview which is located in a provided directory?
I.e. is it possible to set the following attributes (https://docs.asciidoctor.org/asciidoc/latest/attributes/document-attributes-ref/#source-highlighting-and-formatting-attributes):
```
:highlightjsdir: highlight
:highlightjs-theme: vscode_light+
```
where `highlight` is a folder in the current VS Code workspace and `vscode_light+` is a custom theme located in a file `highlight/styles/vscode_light+.min.css`?

My current workaround is to manually copy the file `vscode_light+.min.css` to the VSCode extension folder `.vscode\extensions\asciidoctor.asciidoctor-vscode-2.8.10\media\highlightjs\styles`. Another workaround is to use a similar theme (e.g. `xcode`) only in the preview (i.e. via setting `asciidoc.preview.attributes`) which is already in the `styles` folder of the VS Code extension. 

This is not a big problem for me, but perhaps there is an easy solution?

Mogztter:
@stenzengel This feature is not supported and it might be challenging to implement from a security point of view because you will need to whitelist or explicitly trust this directory.

Please open a new issue so we can discuss this further.

Issue #436, opened 2021-09-08
image preview works in some workspaces, but not in other


aisbergde:
## Description

in some workspaces the the image preview works, in others not. But I can't find differences in workspace settings.

## System Information

Version: 1.60.0 (user setup)
Commit: e7d7e9a9348e6a8cc8c03f877d39cb72e5dfb1ff
Date: 2021-09-01T10:41:52.311Z
Electron: 13.1.8
Chrome: 91.0.4472.164
Node.js: 14.16.0
V8: 9.1.269.39-electron.0
OS: Windows_NT x64 10.0.19042

## To Reproduce

Steps to reproduce the issue:

3. Etc...

## Screenshots & Files

preview working in one workspace

![image](https://user-images.githubusercontent.com/9020187/132517990-7faad8e6-ef80-4669-a0a2-1b716cfec894.png)

not working in another workspace

![image](https://user-images.githubusercontent.com/9020187/132518277-fd6b1f8f-69a2-45c4-9270-6dc4262918e8.png)

but I can open the same file in idea and it's asciidoctor plugin, and it is shown there:

![image](https://user-images.githubusercontent.com/9020187/132518729-9894aa80-93d5-44ac-9ef9-7c4f00539dd4.png)

## Additional Context

there are no different extension settings in these workspaces. But there are some differences in settings for some other extensions. The working folder has:

```
{
  "python.pythonPath": "C:\\Python38\\python.exe",
  "frontMatter.taxonomy.tags": [
    "Azure",
    "DataHandwerk",
 --skipped--
 ],
  "frontMatter.taxonomy.categories": [],
  "pasteImage.path": "${currentFileDir}\\.attachments"
}
```

and the not working folder
```
{
  "pasteImage.path": "${currentFileDir}\\.attachments"
}
```

danyill:
I think I've noticed on Windows that paths with spaces in them and (perhaps) network shares don't work. Are either of these the case?

ajw1970:
I'm having this problem on Windows and I don't have spaces in my path. 

![image](https://user-images.githubusercontent.com/3105267/135131908-8213ae0c-45f2-4702-a8ff-8126df7f55c8.png)

ajw1970:
Version: 1.60.2 (user setup)
Commit: 7f6ab5485bbc008386c4386d08766667e155244e
Date: 2021-09-22T12:00:31.514Z
Electron: 13.1.8
Chrome: 91.0.4472.164
[...]

PR #455, opened 2021-09-29
Restore local file links in preview and support visible editors for source

### LINKED ISSUE #435: Clicking links to local files inside a preview window does not work

danyill:
(closes #435)

### DIFF
--- a/src-preview/index.ts
+++ b/src-preview/index.ts
@@ -130,22 +130,7 @@ document.addEventListener('click', (event) => {
   let node: any = event.target
   while (node) {
     if (node.tagName && node.tagName === 'A' && node.href) {
-      // if (node.getAttribute('href').startsWith('#')) {
-      //   // const fragment = node.href.split('#')[1]
-      //   // if (fragment) {
-      //   //   location.hash = '#' + decodeURI(fragment)
-      //   // }
-      // }
-      // Like VSCode Markdown extension, pass through some known schemes
       let hrefText = node.getAttribute('data-href')
-      if (!hrefText) {
-        // Pass through known schemes
-        if (passThroughLinkSchemes.some((scheme) => node.href.startsWith(scheme))) {
-          return
-        }
-        hrefText = node.getAttribute('href')
-      }
-
       // If original link doesn't look like a url, delegate back to VS Code to resolve
       if (!/^[a-z-]+:/i.test(hrefText)) {
         messaging.postMessage('clickLink', { href: hrefText })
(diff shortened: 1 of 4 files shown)


danyill:
Identifying the resource by the string `'https://file+.vscode-resource.vscode-webview.net/'` may not be a great solution. I can investigate further if this seems too hacky.
I don't think Asciidoctor.js allows a generic link replacement template or an extension to process links.

src-preview/index.ts
```diff
@@ -130,37 +130,32 @@ document.addEventListener('click', (event) => {
   let node: any = event.target
   while (node) {
     if (node.tagName && node.tagName === 'A' && node.href) {
-      // if (node.getAttribute('href').startsWith('#')) {
```
Mogztter, line 133: The Markdown extension is using a "simpler" implementation: https://github.com/microsoft/vscode/blob/main/extensions/markdown-language-features/preview-src/index.ts#L172-L205

Do we need to check if the href starts with `https://file+.vscode-resource.vscode-webview.net/`?

danyill:
I think we need to normalize or rewrite the links for this.

https://github.com/microsoft/vscode/blob/a4342f3addc29f405a5cb4aab359b0b98223411e/extensions/markdown-language-features/src/markdownEngine.ts#L120

https://github.com/microsoft/vscode/blob/a4342f3addc29f405a5cb4aab359b0b98223411e/extensions/markdown-language-features/src/markdownEngine.ts#L257

AFAICT Asciidoctor doesn't yet provide this level of interface (would
require inline parser I think 🤔)

danyill:
Wait up, we should be able to do with a template converter for an inline
anchor?

https://docs.asciidoctor.org/asciidoctor.js/latest/extend/converter/template-converter/

Mogztter:
Yes we can use a custom converter based on the built-in HTML 5 converter to convert links differently.

However, we should fully understand what needs to be done. The Markdown extension does a lot of things, it "normalizes" links at conversion time but then does additional work when a link is clicked (on the WebView).

Maybe links are rewritten in the WebView for security reasons (unless they are using a special schema / domain)?

I find it a bit odd that relative paths (within the same directory) are not automatically handled by VS Code WebView. For instance:

```html
<a href="./first-chapter.adoc">First chapter</a>
```

In my opinion, this should work out-of-the-box when the current working directory is trusted.

Anyway, let's try to find a somewhat elegant solution 😅

danyill:
> Maybe links are rewritten in the WebView for security reasons (unless they are using a special schema / domain)?

Perhaps this blog entry may be useful:

https://blog.mattbierner.com/vscode-webview-web-learnings/

> Anyway, let's try to find a somewhat elegant solution

OK, I'm happy to work on it. We need to engage more constructively with the security model of VS Code to get good outcomes and the continual hacks are counterproductive. But this would be a quick way to restore the existing functionality.

I think the same problems exist with image references as well as links (see #436, I have yet to put my notes on this but I've done a little exploration) so using a custom converter may assist with both of these.

If you're happy let's close this PR and start thinking about a better approach.

Mogztter:
I think we should describe what needs to be done in the issue.

The Markdown extension is probably a good example since it's maintained

PR #454, opened 2021-09-29
Issue #444 - Run Kroki extension within class

danyill:
Closes #444 
This is a rebase of #445.

I have noticed that there appears to be a problem with the highlightjs adapter while testing the rebase of this.

1. With wkhtmltopdf used for PDF generation
2. With `use_kroki` checked try the document below
3. When I export to PDF I receive the following intriguing error:

```bash
[2021-09-29 21:04:08.966] [exthost] [error] TypeError: superclass mismatch for class CustomHighlightJsAdapter
	at ensureSuperclassMatch (/home/mulhollandd/Documents/asciidoctor-vscode/node_modules/asciidoctor-opal-runtime/src/opal.js:479:28)
	at Object.klass (/home/mulhollandd/Documents/asciidoctor-vscode/node_modules/asciidoctor-opal-runtime/src/opal.js:505:9)
	at Object.register (/home/mulhollandd/Documents/asciidoctor-vscode/dist/src/highlightjs-adapter.js:8:43)
	at new AsciidocParser (/home/mulhollandd/Documents/asciidoctor-vscode/dist/src/asciidocParser.js:30:28)
	at ExportAsPDF.<anonymous> (/home/mulhollandd/Documents/asciidoctor-vscode/dist/src/commands/exportAsPDF.js:83:32)
	at Generator.next (<anonymous>)
	at /home/mulhollandd/Documents/asciidoctor-vscode/dist/src/commands/exportAsPDF.js:8:71
	at new Promise (<anonymous>)
	at /home/mulhollandd/Documents/asciidoctor-vscode/dist/src/commands/exportAsPDF.js:4:12
	at ExportAsPDF.execute (/home/mulhollandd/Documents/asciidoctor-vscode/dist/src/commands/exportAsPDF.js:31:16)
	at _executeContributedCommand (/usr/share/code/resources/app/out/vs/workbench/services/extensions/node/extensionHostProcess.js:94:110879)
	at _.$executeContributedCommand (/usr/share/code/resources/app/out/vs/workbench/services/extensions/node/extensionHostProcess.js:94:111317)
	at i._doInvokeHandler (/usr/share/code/resources/app/out/vs/workbench/services/extensions/node/extensionHostProcess.js:98:13543)
	at i._invokeHandler (/usr/share/code/resources/app/out/vs/workbench/services/extensions/node/extensionHostProcess.js:98:13227)
[...]

PR #461, opened 2021-10-03
Fix error in variable naming in spawn process introduced in #454

danyill:
Responds to:

https://github.com/asciidoctor/asciidoctor-vscode/pull/454#discussion_r720812750

### DIFF
--- a/src/asciidocParser.ts
+++ b/src/asciidocParser.ts
@@ -16,7 +16,8 @@ export class AsciidocParser {
   private stylesdir = path.join(this.extPath, 'media')
 
   constructor (private readonly filename: string, private errorCollection: vscode.DiagnosticCollection = null) {
-    this.processor = asciidoctor()
+    this.filename = filename
+    this.errorCollection = errorCollection
   }
 
   public getAttribute (name: string) {
@@ -52,6 +53,7 @@ export class AsciidocParser {
         this.errorCollection.clear()
       }
 
+      this.processor = asciidoctor()
       const memoryLogger = this.processor.MemoryLogger.create()
       this.processor.LoggerManager.setLogger(memoryLogger)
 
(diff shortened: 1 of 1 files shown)


src/asciidocParser.ts
```diff
@@ -16,7 +16,8 @@ export class AsciidocParser {
   private stylesdir = path.join(this.extPath, 'media')
 
   constructor (private readonly filename: string, private errorCollection: vscode.DiagnosticCollection = null) {
-    this.processor = asciidoctor()
+    this.filename = filename
```
danyill, line 19: I put `this.filename` and `this.errorCollection` to avoid a useless constructor lint message -- is this OK?

Mogztter: Last time I ran the linter, I didn't have a warning message.


```typescript
constructor (private readonly filename: string, private errorCollection: vscode.DiagnosticCollection = null) {
```

Using `private readonly filename: string` is valid TypeScript so, technically, we don't need to add `this.filename = filename`.

I wouldn't add extra code "just" to comply with the linter (which apparently does not recognize that this code is valid).In this case, I think it's better to ignore the linter warning.

src/asciidocParser.ts
```diff
@@ -52,6 +53,7 @@ export class AsciidocParser {
         this.errorCollection.clear()
       }
 
+      this.processor = asciidoctor()
```
Mogztter, line 56: We should declare `const processor = asciidoctor()` at the top of this file and replace `this.processor` by `processor`.

src/asciidocParser.ts
```diff
@@ -286,9 +288,9 @@ export class AsciidocParser {
       adocCmdArgs.push('-a', 'env-vscode')
 
       adocCmdArgs.push('-q', '-B', '"' + baseDir + '"', '-o', '-', '-')
-      spawn(adocCmd, adocCmdArgs, options)
+      const asciidoctorProcess = spawn(adocCmd, adocCmdArgs, options)
```
Mogztter, line 291: 👍🏻

Mogztter (changes requested):

src/asciidocParser.ts
danyill: OK, I have taken the long road home...

src/asciidocParser.ts
danyill: Fair enough, however I've done this elsewhere as well... :disappointed: 
I've added a commit to include https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-useless-constructor.md

Mogztter: > I've added a commit to include https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-useless-constructor.md

Yes, that's the right way to fix it 👍

Issue #425, opened 2021-08-05
Does asciidoc have issues with xref and include?


Ormesome:
## Description

Given:
```
:revnumber: v0.1
:revdate: 16.09.2020
:revremark: First draft
:imagesdir: ./img
:toc: top
:icons: font
:experimental: true
:relfilesuffix: .adoc
:relfileprefix: ./chapters/

== Links

image::logo_docker.png[Docker]
xref:docker.adoc[docker]

xref:aws.adoc[aws]

image::logos/logo_azure.png[Azure]
xref:azure.adoc[azure]

xref:infra/ports.adoc[ports]

== Training

include::jobs.adoc[]

include::training/javascript/style.adoc[]

link:https://javascript.info[the javascript training page]
```

When using the preview pane of vscode:

- Both images are fine.
- The xref do not work. Instead, when the xref is clicked on the browser will open with the error: _Check if there is a typo in file+.vscode-resource.vscode-webview.net._
- The include without pathing works.
- The include with pathing does not work. Instead, the preview pane shows: _Unresolved directive in <stdin>_ 
- The link works.

## System Information

Using

- windows 10 home edition.
- vscode 1.58.2, user setup
- asciidoc 2.8.9

## Additional Context

I really want to use asciidoc as a local document repo, a styled way to store tons of notes with tables and images in a hierarchal file structure with the potential to export as HTML.

Is my markup incorrect or does asciidoc have issues with xref and include?

danyill:
Thank you for filing this issue and providing a complete example of the issue. :+1: 

This extension currently seems to have issues with the include preprocessor directive and I suspect with the xref directive also.

I don't see any issue with your markup.
I am also finding difficulty with relative document includes on Linux.

Somehow I think we are not providing the correct document directory/folder when we called asciidoctor.js. I will look into it.

ViToni:
Still valid. Seems to be only `xref` related, using `link` (for local files) instead of `xref` works just fine in the preview pane.
The editor itself can cope with both (Cmd+click). Saving this file 
```
= Document A

xref:document-B.adoc[Document B - Xref]

link:document-B.adoc[Document B - link]
```

creates identical HTML:

```
<div id="content">
<div class="paragraph">
<p><a href="document-B.html">Document B - Xref</a></p>
</div>
<div class="paragraph">
<p><a href="document-B.adoc">Document B - link</a></p>
</div>
```

so it very much looks like the preview pane behaves differently than saving to HTML.

[...]

PR #468, opened 2021-10-17
Part of #425, fix base directory references

danyill:
We recently broken the base directory calling of the processor. I thought it was in e8612434 but that seems to have been just copying across the previous error.

This PR corrects the base directory to match the docs and fixes includes and very likely other things.

https://docs.asciidoctor.org/asciidoctor.js/latest/processor/convert-options/

### DIFF
--- a/src/asciidocParser.ts
+++ b/src/asciidocParser.ts
@@ -119,7 +119,7 @@ export class AsciidocParser {
         attributes: attributes,
         header_footer: true,
         to_file: false,
-        baseDir: baseDir,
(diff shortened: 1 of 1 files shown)


danyill:
I'm fairly confident this is righteous so I might just merge it (famous last words) :stuck_out_tongue_winking_eye: 
I have also identified the criminal! :selfie: in cd477c734561e0b7c966fe62c9c193322dff9f7c
Must remember to use change symbol in future and not find and replace...

Issue #429, opened 2021-08-16
Extension cannot be installed in vs-code web


AlexCzar:
## Description

The extension is not available for installation in vs-code web (e.g. in GitHub web editor)

## System Information

GitHub web-editor (vs-code web)

## To Reproduce

1. Open any repository on github
2. press `.` (dot)
3. when vscode editor opens go to extensions (Ctrl+Shift+X/Cmd+Shift+X)
4. type 'asciidoctor' you'll see the extension grayed out (not-installable)

## Screenshots & Files
![image](https://user-images.githubusercontent.com/1010298/129588015-053e7b75-10da-4b75-9de5-fc178790d70f.png)

## Additional Context

Could be related to #317

Mogztter:
> Could be related to #317

Not sure but I found this page:

> Why is an extension not installable in the browser
There are a small number of extensions that have built-in assumptions or need to run on the desktop. **Examples are when an extension accesses files from the VS Code installation on the desktop or when an extension depends on an executable that must run in a desktop environment.** When you try to install such an extension in the browser, you will be informed that the extension is not available.
Notice such an extension can still be used when connecting to a Codespace from VS Code running on the desktop.

https://code.visualstudio.com/docs/remote/codespaces#_why-is-an-extension-not-installable-in-the-browser

We are using the `child_process`, `fs` and `path` modules (from Node) that might be the reason why the extension is not available.

Mogztter:
It's a bit unfortunate that Visual Code Web does not tell us exactly why the extension is disabled... or provide a tool to check if the extension is compatible or not.

kaovilai:
@Mogztter you can now simply press `.` on the github repo to launch VS Code Web albeit without certain features but you can certainly test out for compatible extensions.

For this repository it would launch web editor if you go to https://github.dev/asciidoctor/asciidoctor-vscode
You should then be able to see installable extensions there.

Mogztter:
> you can now simply press . on the github repo to launch VS Code Web albeit without certain features but you can certainly test out for compatible extensions.
> For this repository it would launch web editor if you go to github.dev/asciidoctor/asciidoctor-vscode
You should then be able to see installable extensions there.

I know but it does not say *why* the extension cannot be installed:

![vscodeweb](https://user-images.githubusercontent.com/333276/133208903-606f3d3d-df05-4481-804b-99b6d9160b22.png)

[...]

PR #473, opened 2021-10-29
Initial work to make the extension available as a Web Extension

Mogztter:
preview

#### What I Did

- Upgrade Webpack from 4.x to 5.60.0 to be able to use a similar configuration to the Markdown extension
- Add a new npm script to build the Web Extension: `npm run build-web`
- Replace `tmp` module by a custom implementation based on `uuid` and `os.tmpdir()`. The reason is that the `tmp` module is using `process.binding` (for some reasons) which is unresolved when the code is loading. As a result VS code was unable to load the extension (packaged with Webpack).
- Replace `extensionPath` by `extensionUri` in `AsciidocContributions`
- Use `vscode.Uri.joinPath` to join paths
- Load the `follow-redirects` module on-demand (similarly to the `tmp` module, this module cannot be loaded correctly in a browser environment)
- Remove `util.promisify`, again it was causing issues with Webpack
- Add `allow-uri-read` and `!copycss: false` attributes to embed the stylesheet from a URI

#### How to test

```
npm run build-web && npx @vscode/test-web --browserType=chromium --extensionDevelopmentPath=${PWD}
```

ref #429


extension-browser.webpack.config.js
```diff
@@ -0,0 +1,79 @@
...
+    // all output goes into `dist`.
+    // packaging depends on that and this must always be like it
+    filename: '[name].js',
+    path: path.join(__dirname, 'dist', 'browser'),
+    libraryTarget: 'commonjs',
+  },
+  amd: false, // disable amd
+  // yes, really source maps
+  devtool: 'source-map',
+  plugins: [
```
Mogztter, line 68: We might want to add `IgnorePlugin` to reduce the size of the bundle: https://webpack.js.org/plugins/ignore-plugin/

package.json
```diff
@@ -71,24 +72,25 @@
 				"icon": {
 					"light": "./media/Preview.svg",
 					"dark": "./media/Preview_inverse.svg"
-				}
+				},
+				"when": "!virtualWorkspace && shellExecutionSupported"
```
Mogztter, line 76: Not sure if this is working... ideally, this command should not be available in a browser environment (because the current implementation relies on a binary `wkhtmltopdf`)

src/asciidocParser.ts
```diff
@@ -126,6 +137,7 @@ export class AsciidocParser {
         backend: backend,
         extension_registry: registry,
       }
+      console.log(options)
```
Mogztter, line 140:
```suggestion
```

Mogztter:
Not perfect but I think this is a good first step on which we can iterate in order to improve the experience when using this extension on VS Code for the Web.

danyill:
I plan to have a look and learn something later this week. How exciting :tada:

danyill:
I didn't really succeed, so far -- I'll try again in the morning:

```bash
$ npm run build-web && npx @vscode/test-web --browserType=chromium --extensionDevelopmentPath=${PWD}

> asciidoctor-vscode@2.9.0-dev build-web /home/mulhollandd/Documents/asciidoctor-vscode
> webpack --config extension-browser.webpack.config --mode none

(node:491526) [DEP_WEBPACK_MODULE_ERRORS] DeprecationWarning: Module.errors was removed (use getErrors instead)
(Use `node --trace-deprecation ...` to show where the warning was created)
asset extension.js 4 MiB [emitted] [big] (name: extension) 1 related asset
runtime modules 1010 bytes 5 modules
modules by path ./node_modules/ 3.81 MiB 211 modules
modules by path ./src/ 172 KiB
  modules by path ./src/*.ts 63.5 KiB 12 modules
  modules by path ./src/commands/*.ts 37.4 KiB 12 modules
  modules by path ./src/features/*.ts 49.5 KiB 8 modules
  modules by path ./src/util/*.ts 9.42 KiB 6 modules
  modules by path ./src/providers/*.ts 11.7 KiB 5 modules
34 modules

WARNING in ./node_modules/asciidoctor-opal-runtime/src/nodejs.js 14:48-55
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
 @ ./node_modules/asciidoctor-opal-runtime/src/index.js 24:2-24
 @ ./src/highlightjs-adapter.ts 4:17-52
 @ ./src/asciidocParser.ts 22:27-59
 @ ./src/asciidocEngine.ts 17:25-52
 @ ./src/extension.ts 17:25-52

1 warning has detailed information that is not shown.
Use 'stats.errorDetails: true' resp. '--stats-error-details' to show it.

webpack 5.60.0 compiled with 1 warning in 4784 ms
Downloading Playwright build of chromium v907428 - 126 Mb [====================] 100% 0.0s 
Downloading Playwright build of ffmpeg v1005 - 2.6 Mb [====================] 100% 0.0s 
Downloading Playwright build of firefox v1281 - 73.7 Mb [====================] 100% 0.0s 
Downloading Playwright build of webkit v1530 - 77.6 Mb [====================] 100% 0.0s 
@vscode/test-web: 0.0.15
Downloading VS Code Insiders: complete
Unpacking VS Code Insiders: complete
Serving dev extensions from /home/mulhollandd/Documents/asciidoctor-vscode
Listening on http://localhost:3000
GET / 200 6.668 ms - 2462
```

And then in the browser I didn't receive a preview, so I looked at the developer tools and saw:

image

danyill:
Ah, this error occurs if I start without saving a file first. Now I have it working. Hurrah, I shall explore more.

I think it dislikes this line in `convertUsingJavascript`:

```js
      const documentPath = path.dirname(path.resolve(doc.fileName))
```

Because this doesn't exist before the file is saved. This doesn't seem to be a problem in desktop applications, I suspect because `path.dirname` or `path.resolve` call a `cwd` function which may not be defined in browser applications but probably becomes something on a filesystem.

Mogztter:
> Because this doesn't exist before the file is saved. This doesn't seem to be a problem in desktop applications, I suspect because path.dirname or path.resolve call a cwd function which may not be defined in browser applications but probably becomes something on a filesystem.

Indeed, Node.js `path` should be replaced by `vscode.Uri` (if possible) or we should revisit the logic.
However, the error in your screenshot mentions `process`, so it might be related to:

https://github.com/asciidoctor/asciidoctor-vscode/blob/283b538aeb9b6d97c2f6aa03601a768f1c16b8f1/src/commands/saveHTML.ts#L41

Please note that `process.env` is defined: https://github.com/asciidoctor/asciidoctor-vscode/pull/473/files#diff-d8349dbc3535b2eb054e38e730dde6244bbfa33b0bc14b1bfe977e991727c418R75

danyill:
> 
> Indeed, Node.js path should be replaced by vscode.Uri (if possible) or we should revisit the logic.
> However, the error in your screenshot mentions process, so it might be related to:
>

I didn't try to save the HTML, so I think the error is (I think) coming from the `resolve` method, which in the browser looks like:

image

Mogztter:
> I didn't try to save the HTML, so I think the error is (I think) coming from the resolve method, which in the browser looks like:

You're right, I didn't know that `path-browserify` was using `process`: https://github.com/browserify/path-browserify/issues/8
Anyway, we will probably need to revisit our path resolution (`base_dir`, `styledir`...) strategy in a browser environment.

As far as I remember, Asciidoctor.js will try its best when running in a browser environment (and when `base_dir` is not set).

I think we should disable all the features related to "styledir" and let "base_dir" undefined (when `process.env.BROWSER_ENV` is true).

danyill:
This is great :tada: :balloon: :cake: :clown_face: :smiley_cat: 
Thank you so much for this.
I've mainly had a little play and made a list of what works and what doesn't.
I would like to read a bit more about some of this but in the meantime...

### Things which are working nicely

* Document symbols
* Preview opens
* Font Awesome integration
* Internal document links
* External document links
* Error messages for problems with the document
* Highlight.js if security is disabled
* Latex Math if security is disabled

### Things I noticed which we need to work on

* [ ] Kroki support
* [ ] CSP prevents MathJax from loading, no equations (similar to existing desktop performance I think)
* [ ] `index.js` is not found (I think this is `index.js` for the `preview-src`) so no communications from Webview 
* [ ] No scroll-sync or communication from webview to main process (because files appear to not be loading)
* [ ] Currently no preview without a file already saved as discussed in the [PR](https://github.com/asciidoctor/asciidoctor-vscode/pull/473#issuecomment-959813236)
* [ ] Images don't appear with either the inline or block macro form
* [ ] Inbuilt highlight.js doesn't appear to be working 
* [ ] Sometimes after the preview is closed it refuses to come back when we try to open it again (only way to get it back appears to be to restart the webserver -- not clear why this might be)
* [ ] Pasting images (uses `process`)
* [ ] Saving to HTML (error no such file: `css/asciidoctor.css`) or exporting to docbook (unknown error, `RunTime error`) or exporting to PDF (`process is not defined`) but these are to be expected

### What do we need for now?

I think we should try for a minimum implementation for which we are just about there = syntax highlighting + a preview of any kind as this gives quite a lot of functionality for anyone pressing `.` on Github.

It will also allow us to see how significant uptake is and how desirable development of this is compared to desktop functionality.

I think if we could get the `preview-src` information to load and images to appear I would say, let's just publish immediately.

### In the future?

Perhaps we could use asciidoctor-web-pdf to open a new window to prepare the PDF.

### Who and When?

Are you happy to work on this? I'd like to finish off some more `2.8.11` issues because we have quite a large backlog of important feature requests but I'm keen to do some work in this area too.

danyill:
Just to say I'll try to look at this in more detail over the next few days. I'd like to try and get the preview script to function.

src/features/previewContentProvider.ts
```diff
@@ -86,22 +87,22 @@ export class AsciidocContentProvider {
...
       </html>`
   }
 
-  private extensionScriptPath (mediaFile: string): string {
-    return vscode.Uri.file(this.context.asAbsolutePath(path.join('dist', mediaFile)))
-      .with({ scheme: 'vscode-resource' })
-      .toString()
+  private extensionResourcePath (webview: Webview, mediaFile: string): string {
+    const webviewResource = webview.asWebviewUri(
+      vscode.Uri.joinPath(this.context.extensionUri, 'media', mediaFile))
```
danyill, line 104: This code expects `index.js` and `pre.js` to be in the `media` directory, but we do not place them there.
My recollection is that I introduced this distinction between resources and scripts so that we could webpack the preview into a different folder.

If I manually place these files into this folder then the preview scrolling works.

I think we do want these to be in the 'dist' folder somehow because they are built artifacts.

danyill, line 104: Oh the solution appears to be happily straightforward:

```suggestion
      vscode.Uri.joinPath(this.context.extensionUri, 'dist', mediaFile))
```

package.json
danyill: I thought we could just use `!isWeb`? Alas, it did not work.

The docs said:


I also tried `!virtualWorkspace` (noting [this](https://github.com/microsoft/vscode/issues/121165) might provide a clue why it is not working), or perhaps [this](https://github.com/microsoft/vscode/issues/124594) which might suggest it can't be used for debugging.

But presumably on the desktop client we could still have a virtualWorkspace? See https://code.visualstudio.com/api/extension-guides/virtual-workspaces.

I thought !isWeb 

I can't see a solution at this time.

danyill: Finally, I tried:

`"workspacePlatform != webworker"` which also did not work... :pouting_cat: 

When I use the [Inspect Context Keys](
https://code.visualstudio.com/api/references/when-clause-contexts#inspect-context-keys-utility) utility, I can find some interesting variables:

In the web browser running on Linux.

```js
isLinux: true
isMac: false
isMacNative: false
isWeb: true
isWindows: false
virtualWorkspace: ""
workspacePlatform: "webworker"
```

`shellExecutionSupported` doesn't appear to exist at all.

Comparing with local installation running in vs code on Linux:

```js
isLinux: true
isMac: false
isMacNative: false
isWeb: false
isWindows: false
shellExecutionSupported: true
virtualWorkspace: ""
workspacePlatform: "linux"
```

I think the debug environment for the web extension perhaps doesn't support this for some reason. 

I think we should use:

```js
"when": "!web",
```
on the following commands initially to support a MVP:

```js
"command": "asciidoc.exportAsPDF",
"command": "asciidoc.pasteImage",
"command": "asciidoc.saveHTML",
"command": "asciidoc.saveDocbook",
```

Perhaps we'll have to do a trial release to verify this.

extension-browser.webpack.config.js
danyill: Alas, I do not understand how we could use this.

`extension.js` seems to be about 4 MiB and  `extension.js.map` which presumably we should only produce for development is about 4.8 MiB on my machine.

danyill (changes requested):
I've made a few comments and learned a few things. Thanks for this opportunity to walk together :walking_man:  :walking_man: 

It looks straightforward to bring back communications with the WebView and we must do this because it would currently break the conventional build as well.

I suspect reinstating images is the next tricky thing. I have not looked into this but I suspect we need to make their paths more "URI" like, possibly with a custom converter?

Perhaps we should clarify in the README any features which we don't presently provide in the web environment?

In my running of this PR code in the conventional arrangement, the changes seemed to work fine in general as far as I can tell.

Incidentally, having a new target platform does increase the need in my mind to create some automated tests. To detect breakages because testing across four platforms is onerous and error prone and often there are tricky requirements between Windows/Linux and now some differences now required with the web applications as well. I'll file an issue for this before long.

Anyway this is a super-cool :sunglasses: feature - thanks for this great work.

src/features/previewContentProvider.ts
Mogztter: (just lost my comment by clicking "unresolve conversation" 😠)

> My recollection is that I introduced this distinction between resources and scripts so that we could webpack the preview into a different folder.

Could you please tell me more about this choice? I don't see any benefit of using a different directory.
For reference, the Markdown extension is also using Webpack with the `media` directory as the `output`. They are also using a `.gitignore` to ignore the generated JavaScript files:

- `webpack.config.js`: https://github.com/microsoft/vscode/blob/66b1668b66768275b655cde14d96203915feca7b/extensions/markdown-language-features/webpack.config.js#L28
- `.gitignore`: https://github.com/microsoft/vscode/blob/66b1668b66768275b655cde14d96203915feca7b/extensions/markdown-language-features/.gitignore#L2

Using the same directory means that "all" webview resources are in the same place. We can then use the same function to resolve the path to `index.js`, `pre.js`, `highlight.css`, `asciidoctor-default.css`...

extension-browser.webpack.config.js
Mogztter: We are using a lot of polyfills and some features won't be available in a web environment but the code will still be included in the bundle.
My idea is to use the `IgnorePlugin` to make sure that we only include what we need.
But we can do that at a later point.

Mogztter:
> [ ] Sometimes after the preview is closed it refuses to come back when we try to open it again (only way to get it back appears to be to restart the webserver -- not clear why this might be)

I also noticed that but I hope this is just an issue with `@vscode/test-web`...

> I suspect reinstating images is the next tricky thing. I have not looked into this but I suspect we need to make their paths more "URI" like, possibly with a custom converter?

Probably, I will test.

> Perhaps we should clarify in the README any features which we don't presently provide in the web environment?

Yes, that's a good idea. We could even do a feature matrix table:
|Feature|Desktop|Web|
|--|--|--|
|Preview HTML|✔️|✔️|
|Sync scrolling between the editor and the preview|✔️|✔️|
|Save as HTML|✔️|⛔|
|Export as PDF|✔️|⛔|
|...|...|...|

danyill:
> Using the same directory means that "all" webview resources are in the
same place. We can then use the same function to resolve the path to
index.js, pre.js, highlight.css, asciidoctor-default.css...

I think perhaps it's better to follow "upstream" as you suggest.

My reasons were mainly about separating built and generated artifacts in
the taxonomy but I'm not very attached to it (although I would like a good
clean command if we are going to integrate with git and not separate output
in a directory).

package.json
Mogztter: Done!

src/features/previewContentProvider.ts
Mogztter: Let's create a follow-up issue.

Mogztter:
Not sure why but when the developer tools are opened the preview does not work 😞 

Also, when the preview tab is not active/visible, the preview won't reactivate. In the console, I get:

```
Uncaught (in promise) DOMException: Failed to execute 'postMessage' on 'Client': Port at index 0 is already neutered.
```

Might be related to: https://github.com/microsoft/vscode/issues/139346

I think we should publish an alpha version to test on https://insiders.vscode.dev/

Mogztter:
>  Images don't appear with either the inline or block macro form

Could you please provide a reproduction case? Seems to be working fine on my end:

image

Mogztter:
> CSP prevents MathJax from loading, no equations (similar to existing desktop performance I think)

That's correct, the behavior is the same. If we want to address this issue we should create a follow-up issue.

Mogztter:
> Kroki support

It's working with the latest version published 🎉 

image

danyill:
Sorry, I've been a bit distracted, I'll take a look in the next day or two.

danyill:
When compiling I get three errors:

```
WARNING in ./node_modules/asciidoctor-opal-runtime/src/nodejs.js 14:48-55
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
 @ ./node_modules/asciidoctor-opal-runtime/src/index.js 24:2-24
 @ ./src/highlightjs-adapter.ts 4:17-52
 @ ./src/asciidocParser.ts 22:27-59
 @ ./src/asciidocEngine.ts 17:25-52
 @ ./src/extension.ts 17:25-52

WARNING in ./node_modules/asciidoctor-kroki/dist/browser/asciidoctor-kroki.js 1:491-492
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
 @ ./src/asciidocParser.ts 19:14-42
 @ ./src/asciidocEngine.ts 17:25-52
 @ ./src/extension.ts 17:25-52

WARNING in ./node_modules/asciidoctor-kroki/dist/browser/asciidoctor-kroki.js 1:500-501
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
 @ ./src/asciidocParser.ts 19:14-42
 @ ./src/asciidocEngine.ts 17:25-52
 @ ./src/extension.ts 17:25-52

3 warnings have detailed information that is not shown.
Use 'stats.errorDetails: true' resp. '--stats-error-details' to show it.
```

I don't know what this means.

When I open a document "Open File" with images I still receive many errors but I think these are because the files are not being served by a web server. I can't see local directory files. Does that sound right?

Also the following CSP errors for:

* highlight.js
* MathJax

```
Refused to load the stylesheet 'http://localhost:3000/static/devextensions/media/highlightjs/styles/github.min.css' because it violates the following Content Security Policy directive: "style-src https://*.vscode-webview.net 'unsafe-inline' https: data:". Note that 'style-src-elem' was not explicitly set, so 'style-src' is used as a fallback.

index.html:1 Refused to load the script 'http://localhost:3000/static/devextensions/media/highlightjs/highlight.min.js' because it violates the following Content Security Policy directive: "script-src https://*.vscode-webview.net 'nonce-1642293009469469' 'sha256-ZrDBcrmObbqhVV/Mag2fT/y08UJGejdW7UWyEsi4DXw='". Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.

index.html?id=2a26b3f2-db4e-45d6-b124-a6c011672ec4&swVersion=4&extensionId=asciidoctor.asciidoctor-vscode&platform=browser&vscode-resource-base-authority=vscode-resource.vscode-webview.net&parentOrigin=http%3A%2F%2Flocalhost%3A3000:2 Uncaught ReferenceError: hljs is not defined
    at index.html?id=2a26b3f2-db4e-45d6-b124-a6c011672ec4&swVersion=4&extensionId=asciidoctor.asciidoctor-vscode&platform=browser&vscode-resource-base-authority=vscode-resource.vscode-webview.net&parentOrigin=http%3A%2F%2Flocalhost%3A3000:2
(anonymous) @ index.html?id=2a26b3f2-db4e-45d6-b124-a6c011672ec4&swVersion=4&extensionId=asciidoctor.asciidoctor-vscode&platform=browser&vscode-resource-base-authority=vscode-resource.vscode-webview.net&parentOrigin=http%3A%2F%2Flocalhost%3A3000:2
index.html:1 Refused to load the script 'https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.9/MathJax.js?config=TeX-MML-AM_HTMLorMML' because it violates the following Content Security Policy directive: "script-src https://*.vscode-webview.net 'nonce-1642293009469469' 'sha256-ZrDBcrmObbqhVV/Mag2fT/y08UJGejdW7UWyEsi4DXw='". Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.
```

On (perhaps) long documents I'm not seeing source highlighting for headings etc.
For instance https://gitlab.com/danyill/asciidoctor-quick-reference/-/blob/master/asciidoctor-cheat-sheet.adoc

I sometimes see in the output of the "JSON Language Server"

```
[Error - 13:44:24] Connection to server is erroring. Shutting down server.
[Error - 13:44:24] Connection to server is erroring. Shutting down server.
```

I don't think any of this is the end of the world. Can we do a pre-publish or beta to allow testing on Github? That would help us confirm how much of this is a possible artifact of the development environment. Additionally it would allow for instance testing "paste image" where we need to have a "folder open" to test properly (I think).

I think we should do a table like you've shown in https://github.com/asciidoctor/asciidoctor-vscode/pull/473#issuecomment-962484638

How about this as a starter for ten:

|Feature|Desktop|Web|
|--|--|--|
|Document Outline and Symbols|✔️|✔️|
|Equations (via Mathjax)|✔️|⛔|
|Export as PDF|✔️|⛔|
|Kroki Integration for Diagrams|✔️|✔️|
|Paste Image |✔️|⛔|
|Save as HTML|✔️|⛔|
|Save as Docbook|✔️|⛔|
|Snippets|✔️|✔️|
|Syntax Highlighting|✔️|⛔|
|Sync scrolling between the editor and the preview|✔️|✔️|
|Syntax Highlighting|✔️|✔️|

I did some testing for the "non-web" version and tested Kroki which worked fine.

danyill:
Ah, no Syntax Highlighting is fine if Security is disabled.

danyill:
And so are equations

Mogztter:
> When compiling I get three errors:

These are warnings, you should still be able to build the extension.
I'm not entirely sure why, we are using Browserify to produce a browser-compatible version of Asciidoctor Kroki. I guess some constructs are not well-supported by Webpack...

> I think we should do a table like you've shown in #473 (comment)
How about this as a starter for ten:

Sounds good, should I add it in the README?

> Ah, no Syntax Highlighting is fine if Security is disabled.
> And so are equations

👍🏻 

> I don't think any of this is the end of the world. Can we do a pre-publish or beta to allow testing on Github? That would help us confirm how much of this is a possible artifact of the development environment. Additionally it would allow for instance testing "paste image" where we need to have a "folder open" to test properly (I think).

I agree, I think we should publish an alpha/beta.

@danyill How do you want to proceed?

danyill:
We need to have workflow support for pre-release I think. I've filed #488. There's a question on there that I don't have an answer for (how to detect version is odd/even without duplicating jobs) but otherwise it seems straightforward.

danyill (approved):

danyill:
From your comments on #488 then we just need to make a minor tweak to the README and we're good to do a release.
If you're happy to help me knock #474 into shape first I'd be grateful if we could get that out at the same time.

PR #475, opened 2021-10-31
Scroll to the first line when the preview hits top

### LINKED ISSUE #348: Something weird when scrolling?
## Description

1st thing)
If i have a preview running next to an .adoc file and scrolls, having focus on the .adoc file, When scrolling from below and up... the preview will stop at the first level heading while the .adoc will scroll to the top showing the title. (starting with a '='

The top of the document renders and I can reach it by switching focus to the preview and continue the scroll upwards.

2nd thing)
When I switch focus to the preview and scroll to the top (I.e. i see the title, the scroll in the .adoc file does not reach (i.e. show) the top-most row.

## System Information

Version: 1.46.1 (user setup)
Commit: cd9ea6488829f560dc949a8b2fb789f3cdc05f5d
Date: 2020-06-17T21:13:20.174Z
Electron: 7.3.1
Chrome: 78.0.3904.130
Node.js: 12.8.1
V8: 7.8.279.23-electron.0
OS: Windows_NT x64 10.0.19041

Asciidoc 2.8.1

## To Reproduce

Steps to reproduce the issue:

1. Go to https://asciidoctor.org/docs/asciidoc-article/ and copy that asciidoc example (to get something to demo on)
2. Setup with having an .adoc next to an asciidoctor preview (i.e. split window)
3. Click on the adoc file and scroll to the bottom. (the preview tags along).
4. Scroll back up to the top of the .adoc file so you see the 1st row (the 1st row starts with "= AsciiDoc Article Title"
5. The preview does not show the title of the document. Instead it shows 1 row above the first "first level heading" (see screenshot)
6 swap focus to the preview and scroll to the top.
7. You can then see the rendered title but... now you can't see the 1st row of the .adoc file. (screenshot 2)

## Screenshots & Files

Screenshot 1.

image

Screenshot 2.

image

Mogztter:
Since the document title does not have a `data-line` attribute when we scroll to the top on the preview it does not scroll to the document title (but to the first block).

I didn't find a way to do the opposite (i.e., when we scroll to the top of the AsciiDoc document, scroll to the top of the preview). The reason is that we cannot tell what is on the first line. I guess we could read the first line and check if it's the document title but even then the document title could be hidden in the preview with `:showtitle!:`.

So this pull request partially fixes #348

### DIFF
--- a/preview-src/index.ts
+++ b/preview-src/index.ts
@@ -6,7 +6,7 @@ import { ActiveLineMarker } from './activeLineMarker'
 import { onceDocumentLoaded } from './events'
 import { createPosterForVsCode } from './messaging'
 import { getEditorLineNumberForPageOffset, scrollToRevealSourceLine } from './scroll-sync'
-import { getSettings, getData } from './settings'
+import { getData, getSettings } from './settings'
 import throttle = require('lodash.throttle');
 
 declare let acquireVsCodeApi: any
@@ -89,7 +89,6 @@ window.addEventListener('message', (event) => {
     return
   }
 
-  //console.log("GOT MESSAGE", event.data);
   switch (event.data.type) {
     case 'onDidChangeTextEditorSelection':
       marker.onDidChangeTextEditorSelection(event.data.line)
@@ -161,7 +160,10 @@ if (settings.scrollEditorWithPreview) {
       scrollDisabled = false
     } else {
       const line = getEditorLineNumberForPageOffset(window.scrollY)
-      if (typeof line === 'number' && !isNaN(line)) {
+      if (window.scrollY === 0) {
+        // scroll to top, document title does not have a data-line attribute
+        messaging.postMessage('revealLine', { line: 0 })
+      } else if (typeof line === 'number' && !isNaN(line)) {
         messaging.postMessage('revealLine', { line })
       }
     }
--- a/preview-src/scroll-sync.ts
+++ b/preview-src/scroll-sync.ts
@@ -96,11 +96,13 @@ export function scrollToRevealSourceLine (line: number) {
       // Between two elements. Go to percentage offset between them.
       const betweenProgress = (line - previous.line) / (next.line - previous.line)
       const elementOffset = next.element.getBoundingClientRect().top - previousTop
-      scrollTo = previousTop + betweenProgress * elementOffset
+      scrollTo = window.scrollY + previousTop + betweenProgress * elementOffset
+    } else if (line === 0) {
+      scrollTo = 0
     } else {
-      scrollTo = previousTop
+      scrollTo = window.scrollY + previousTop
     }
-    window.scroll(0, Math.max(1, window.scrollY + scrollTo))
+    window.scroll(0, Math.max(1, scrollTo))
   }
 }
 


Mogztter:
> I didn't find a way to do the opposite (i.e., when we scroll to the top of the AsciiDoc document, scroll to the top of the preview). The reason is that we cannot tell what is on the first line. I guess we could read the first line and check if it's the document title but even then the document title could be hidden in the preview with :showtitle!:.

Arguably, we could scroll to top no matter what... I think this is the expected behavior anyway. Thoughts?

preview-src/index.ts
```diff
@@ -161,7 +160,10 @@ if (settings.scrollEditorWithPreview) {
       scrollDisabled = false
     } else {
       const line = getEditorLineNumberForPageOffset(window.scrollY)
-      if (typeof line === 'number' && !isNaN(line)) {
+      if (window.scrollY === 0) {
+        // scroll to top, document title does not have a data-line attribute
+        messaging.postMessage('revealLine', { line: 1 })
```
danyill, line 165: I think lines are indexed beginning at `0` :stuck_out_tongue_winking_eye: 

When I had `1` My source document line stopped at line `2` when I was at the top of the preview.

We gaze upon our tooltip assistance and see it is zero-indexed:

image

danyill:
```suggestion
        messaging.postMessage('revealLine', { line: 0 })
```

danyill (changes requested):
A little niggardly, I admit.

danyill:
> > I didn't find a way to do the opposite (i.e., when we scroll to the top of the AsciiDoc document, scroll to the top of the preview). The reason is that we cannot tell what is on the first line. I guess we could read the first line and check if it's the document title but even then the document title could be hidden in the preview with :showtitle!:.
> 
> Arguably, we could scroll to top no matter what... I think this is the expected behavior anyway. Thoughts?

I am not sure, I probably wouldn't put much effort into it. I don't think the preview and source should be "perfect" because the underlying alignment is not 1-to-1 (we have hidden things, comments, document attributes etc.).

I am concerned about side effects and jumping back and forth for a single line change for large document headers. There is already a problem like this at the _bottom_ of some documents which I've experienced where the preview and source pane flutter alarmingly whenever you scroll below a particular point. On small form factors or when the window isn't maximized this could be an issue, perhaps when you scroll after the title or after the document attributes or to the first `data-line-x` value.

We also have this problem for tables where which cross more than one screen because we can't allocate `data-line-x` on the table cells or rows.

I guess the proof would be to implement it and see if it can be made suitably robust.

Mogztter:
> I am not sure, I probably wouldn't put much effort into it. I don't think the preview and source should be "perfect" because the underlying alignment is not 1-to-1 (we have hidden things, comments, document attributes etc.).

That's true but when the scrollbar hits the top (or the bottom) it should also hits the top (or the bottom) of the preview/editor.

> I am concerned about side effects and jumping back and forth for a single line change for large document headers. There is already a problem like this at the bottom of some documents which I've experienced where the preview and source pane flutter alarmingly whenever you scroll below a particular point. On small form factors or when the window isn't maximized this could be an issue, perhaps when you scroll after the title or after the document attributes or to the first data-line-x value.

Here's the result:

jump

Not sure if it's possible but we could "smooth scroll" instead of "jumping".

In my opinion, today it's even worse since you cannot have both scrollbars at the top (i.e., show the document title on both the preview and the editor):

showtitle

> We also have this problem for tables where which cross more than one screen because we can't allocate data-line-x on the table cells or rows.

As far as I know this improvement does not improve or worsen the situation regarding tables.
We can address this concern but not as part of this pull request since that's another issue.

> I guess the proof would be to implement it and see if it can be made suitably robust.

I did and I think it's better, now both sides will be synced when they reach the top.

preview-src/index.ts
Mogztter: Good catch!

danyill:
> I did and I think it's better, now both sides will be synced when they reach the top.

I agree, thanks for this! Let's do it.

danyill (approved):

PR #494, opened 2022-01-30
Add link provider for includes, closes #99

### LINKED ISSUE #99: Links don't work properly
I might be doing something wrong, but here's an example of a link in one of my documents:

`* https://jvns.ca/networking-zine.pdf[Networking, Ack! Zine]`

In the Asciidoc output it works fine, but inside the document when editing, the link becomes (`https://jvns.ca/networking-zine.pdf[Networking`), so if I click the link to visit a site, I get an error because `[Networking` is not a part of the URL.

danyill:
**What I did:**

*    Used an IncludeProcessor to find the link text and reader line numbers for includes and then regex parsed the document to identify line numbers more precisely since the Asciidoctor reader is "imprecise" (optimised for speed, not accuracy).
* Created a line number alignment algorithm to process include line numbers from the IncludeProcessor and from regex parsing to "match up" the relevant lines.
* Abstracted the link sanitizer code previously only used in the preview as part of AsciidoctorWebViewConverter to be part of the normalizeLinks routine so that preview and DocumentLinks use the same implementation.

**Implementation limitations:**

Items within an `ifdef` or `ifeval` which aren't processed won't receive DocumentLinks because they're not handled by the include processor. But to avoid this would require quite sophisticated external logic in this extension and doesn't currently seem worthwhile.

**What I didn't do:**

Write a test. I'm not entirely sure how but I haven't spent any time on it. I have commented out the tests because it was causing the linter to be argumentative.
I can work on this if you would prefer to have this done prior to merging but I would like to see if we think this PR is heading in the right direction.

I've done some initial testing and the logic seems to work across a range of documents that were close at hand... but some more testing would be appreciated.

_NOTE:_ Originally this PR was part of #474 but this was trying to do inline links as well and this was not possible and I was unable to effectively rebase it so I've taken a diff of the final commit to the originating commit to generate this PR and rebased on the current master.

### DIFF
--- a/src/asciidocParser.ts
+++ b/src/asciidocParser.ts
@@ -2,6 +2,7 @@ import * as vscode from 'vscode'
 import * as path from 'path'
 import { spawn } from 'child_process'
 import { AsciidoctorWebViewConverter } from './asciidoctorWebViewConverter'
+const asciidoctorFindIncludeProcessor = require('./asciidoctorFindIncludeProcessor')
 
 const asciidoctor = require('@asciidoctor/core')
 const docbook = require('@asciidoctor/docbook-converter')
@@ -9,15 +10,16 @@ const kroki = require('asciidoctor-kroki')
 const processor = asciidoctor()
 const highlightjsBuiltInSyntaxHighlighter = processor.SyntaxHighlighter.for('highlight.js')
 const highlightjsAdapter = require('./highlightjs-adapter')
-processor.ConverterFactory.register(new AsciidoctorWebViewConverter(), ['html5'])
 
 export class AsciidocParser {
   public html: string = ''
   public document = null
   public processor = null
-  private stylesdir: string
 
-  constructor (private readonly filename: string, private errorCollection: vscode.DiagnosticCollection = null) {
+  private stylesdir: string
+  public baseDocumentIncludeItems = null
+
+  constructor (public filename: string, private errorCollection: vscode.DiagnosticCollection = null) {
     const extensionContext = vscode.extensions.getExtension('asciidoctor.asciidoctor-vscode')
     this.stylesdir = vscode.Uri.joinPath(extensionContext.extensionUri, 'media').toString()
   }
@@ -31,10 +33,11 @@ export class AsciidocParser {
     return match
   }
 
-  private async convertUsingJavascript (text: string,
+  public async convertUsingJavascript (text: string,
     doc: vscode.TextDocument,
     forHTMLSave: boolean,
     backend: string,
+    getDocumentInformation: boolean,
     context?: vscode.ExtensionContext,
     editor?: vscode.WebviewPanel) {
     return new Promise<string>((resolve, reject) => {
@@ -60,11 +63,21 @@ export class AsciidocParser {
       processor.LoggerManager.setLogger(memoryLogger)
 
       const registry = processor.Extensions.create()
+      // registry for processing document differently to find AST/metadata otherwise not available
+      const registryForDocumentInfo = processor.Extensions.create()
 
+      const asciidoctorWebViewConverter = new AsciidoctorWebViewConverter()
+      processor.ConverterFactory.register(asciidoctorWebViewConverter, ['html5'])
       const useKroki = vscode.workspace.getConfiguration('asciidoc', null).get('use_kroki')
 
       if (useKroki) {
         kroki.register(registry)
+      }
+
+      // the include processor is only run to identify includes, not to process them
+      if (getDocumentInformation) {
+        asciidoctorFindIncludeProcessor.register(registryForDocumentInfo)
+        asciidoctorFindIncludeProcessor.resetIncludes()
       }
 
       if (context && editor) {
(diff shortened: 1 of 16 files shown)


src/asciidoctorFindIncludeProcessor.ts
```diff
@@ -0,0 +1,49 @@
...
+  length: string
+}
+
+interface IncludeItem {
+  [index: number]: IncludeEntry
+}
+
+interface IncludeItems extends Array<IncludeItem>{}
+
+let baseDocIncludes: IncludeItems[] = []
```
Mogztter, line 13: I think it would be better to use an array of `IncludeEntry` rather than an array of `Array<IncludeItem>`

```js
[
  { name: 'a.adoc', position: 5, length: 6 },
  { name: 'github.adoc', position: 10, length: 11 },
  { name: 'windows.adoc', position: 16, length: 12 },
  { name: 'b.adoc', position: 27, length: 6 },
  { name: 'a.adoc', position: 31, length: 6 },
  { name: 'c.adoc', position: 45, length: 6 }
]
```

Mogztter:
> Items within an ifdef or ifeval which aren't processed won't receive DocumentLinks because they're not handled by the include processor. But to avoid this would require quite sophisticated external logic in this extension and doesn't currently seem worthwhile.

I can think of an idea to workaround that but we should discuss it in a follow-up issue/pull request.
Another limitation is that attributes won't be resolved if they are defined in an include.

**main.adoc**
```
include::attrs.adoc[]

include::{operating-system-name}.adoc[]
```

**attrs.adoc**
```
:operating-system: macos
```

Again we should address this limitation in another issue/pull request.

Mogztter:
> Write a test. I'm not entirely sure how but I haven't spent any time on it. I have commented out the tests because it was causing the linter to be argumentative.

On my machine the `npm run lint` task runs fine if I uncomment `documentLinkProvider.test.ts`

src/asciidocParser.ts
```diff
@@ -2,42 +2,37 @@ import * as vscode from 'vscode'
...
+  public baseDocumentIncludeItems = null
 
-  constructor (private readonly filename: string, private errorCollection: vscode.DiagnosticCollection = null) {
-    const extensionContext = vscode.extensions.getExtension('asciidoctor.asciidoctor-vscode')
-    this.stylesdir = vscode.Uri.joinPath(extensionContext.extensionUri, 'media').toString()
-  }
-
-  public getAttribute (name: string) {
-    return (this.document == null) ? null : this.document.getAttribute(name)
+  constructor (extensionUri: vscode.Uri, private errorCollection: vscode.DiagnosticCollection = null) {
```
Mogztter, line 20: This class desperately needs to be rewritten.
We should have dissociate the `load` function from the `convert` function and avoid using mutable class variables.

The `load` and `convert` functions should be pure.

src/image-paste.ts
```diff
@@ -327,10 +327,12 @@ export namespace Import {
       }
 
       if (dir === '') {
-        const thisDocument = vscode.window.activeTextEditor.document
-        const adoc = new AsciidocParser(thisDocument.uri.fsPath)
-        adoc.parseText(thisDocument.getText(), thisDocument)
-        dir = adoc.document.getAttribute('imagesdir')
+        const textDocument = vscode.window.activeTextEditor.document
+        const extensionUri = vscode.Uri.file('') // won't be used anyway... needs refactoring!
+        const {document} = await new AsciidocParser(extensionUri).parseText(textDocument.getText(), textDocument)
```
Mogztter, line 332: Since we don't need to convert the document, we don't need to define a valid `extensionUri`.

Mogztter:
@danyill tests are running fine now, I had to remove `vscode.extensions.getExtension('asciidoctor.asciidoctor-vscode')` (which is, I believe, an anti-pattern anyway).

I'm already much more confident now that we have a few automated tests running. The code coverage is pretty low but that's a start.

Since I made quite a few changes, feel free to manually test to make sure that I didn't break existing features.
I'm ready to merge when you are ready.

In a following pull request I would like to rewrite/dissociate/simplify `AsciidocEngine` and `AsciidocParser`.

danyill:
>  Another limitation is that attributes won't be resolved if they are defined in an include.

Really good point and it didn't occur to me at all.

I think we could perhaps do something like what was suggested on Zulip by Dan here: https://asciidoctor.zulipchat.com/#narrow/stream/279642-users/topic/Adding.20'optional'.20include.3A.3A.20option.20via.20extension/near/261486251

I filed: https://github.com/asciidoctor/asciidoctor-vscode/issues/496

src/asciidocParser.ts
danyill: I'm happy to look at this, perhaps we can discuss on the issue.

src/asciidoctorFindIncludeProcessor.ts
danyill: Yes, it is better. This was my hacky way of getting an index position easily. I will improve.

danyill:
Thank you for this, I am delighted to have some tests which we can build on.

Thank you also for your review comments, I have changed the way IncludeEntries are handled to use a better object definition. I appreciate the tidying of the code base as you make other changes. There is much tidying possible.

I force pushed this because this did not resolve #99, but rather #498 and I wanted to fix the commit history. I added a CHANGELOG entry, hoping to get this into the next release.

If you are happy feel free to merge.

Mogztter:
It looks good to me, nice work 👍

PR #502, opened 2022-02-02
Add option stem_renderer

tani:
Among the several modern asciidoctor applications, the latest version of MathJax is employed.
For example, asciidoctor-reveal.js uses it. Thus, the preview of documentation with the latest is in demand. However, we cannot do that because the viewer uses compiled version of asciidoctor, which still uses the old MathJax for compatibility. This pull request adds an option `stem_renderer` to specify the render for stem notation. Note that I add only two enums `MathJax2` as default and `MathJax3`. In the future, I would like another render, i.e., KaTeX. However, at this moment, I failed for unclear reasons.
Let me add it at another pull request.

Thanks for the great software, I enjoy to use this on vscode.dev.
Cheers.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -639,6 +639,11 @@
 				}
 			}
 		},
+		"asciidoctor-mathjax3": {
+			"version": "0.0.2",
+			"resolved": "https://registry.npmjs.org/asciidoctor-mathjax3/-/asciidoctor-mathjax3-0.0.2.tgz",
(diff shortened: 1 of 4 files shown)


Mogztter:
Thanks for your contribution!

Please open a new issue so we can decide if we want to support this feature or not, if this should be part of this extension (built-int) or available as an add-on.

Mogztter:
Thanks again for your work but as mentioned in https://github.com/asciidoctor/asciidoctor-vscode/issues/505#issuecomment-1162266900 we won't implement it until a pluggable interface is available in Asciidoctor core.

Once this interface is available, feel free to submit a pull request again to add MathJax 3.x support.

PR #506, opened 2022-02-02
Use esbuild to produce a single file

### LINKED ISSUE #491: Use target es2020 and minimum version to 1.56 to reduce bundle size
Reference: https://jason-williams.co.uk/speeding-up-vscode-extensions-in-2022

ggrossetie:
- Upgrade target from es6 to es2020 (to reduce bundle size)
- Upgrade minimum VS code version to 1.60.0

#### TODO

- [ ] make sure that localization is still working when using a single file
 
resolves #491


Mogztter:
Currently, the bundle size is 2.8mb.
I didn't notice any performance improvement on desktop but `esbuild` is pretty fast and hopefully the tree shaking will eventually decrease the bundle size.

We might also want to use `esbuild` to bundle the web extension (in a follow-up pull request) as, currently, Webpack produces a 3.6mb bundle. Alternatively, we could try to tune Webpack. For reference, the current configuration is heavily inspired by: https://github.com/microsoft/vscode/blob/main/extensions/markdown-language-features/extension.webpack.config.js

Issue #501, opened 2022-02-02
VSC preview in 2.9.3 is not working as in 2.8.10


ioannisstavrakakis:
## Description

The VSC preview in v2.9.3 is not working the same way as in v2.8.10.
I use the option:
**Asciidoc › Preview: Use Editor Style**
_Use editor style instead of default asciidoctor.css_

The problem is also seen in v2.9.2, so it is something between this and v2.8.10 that broke the preview.

There are so many changes between the two previews that it is best if you check the attached gif files to see what exactly has been changed.

## System Information

Visual Studio Code:
Version: 1.63.2 (user setup)
Commit: 899d46d82c4c95423fb7e10e68eba52050e30ba3
Date: 2021-12-15T09:40:02.816Z
Electron: 13.5.2
Chromium: 91.0.4472.164
Node.js: 14.16.0
V8: 9.1.269.39-electron.0
OS: Windows_NT x64 10.0.19044

asciidoctor-vscode v2.9.3

## To Reproduce

Steps to reproduce the issue:

1. Go to extension settings.
2. Go to **Asciidoc › Preview: Use Editor Style**
3. Select the checkbox _Use editor style instead of default asciidoctor.css_
4. Check adoc preview pane.

### 2.9.3. preview
![2 9 3](https://user-images.githubusercontent.com/87022642/152146932-5c84ebc9-22dd-4a05-8d1b-c95d21f7b6bc.gif)

### 2.8.10 preview
![2 8 10](https://user-images.githubusercontent.com/87022642/152147574-b1147ee8-2363-4355-a226-38f81678465b.gif)

Mogztter:
Could you please share your document?

ioannisstavrakakis:
It is just a doc with various adoc syntaxes for reference. There you go:
[asciidoc_syntax_reference_guide.zip](https://github.com/asciidoctor/asciidoctor-vscode/files/7986413/asciidoc_syntax_reference_guide.zip)

Mogztter:
I cannot reproduce this issue:

![image](https://user-images.githubusercontent.com/333276/152176470-fec9166a-156f-4362-94e8-142aa5baa592.png)


```
Version: 1.63.2
Commit: 899d46d82c4c95423fb7e10e68eba52050e30ba3
Date: 2021-12-15T09:39:46.686Z
Electron: 13.5.2
Chromium: 91.0.4472.164
Node.js: 14.16.0
V8: 9.1.269.39-electron.0
[...]

PR #510, opened 2022-02-05
Issue 501 preview style regression

danyill:
See #501 for discussion. I will be happy to take a review.

I'm a little surprised the underlying VS Code API doesn't handle this more elegantly but I haven't found a way to do this other than to check the environment.

I've decided to retain the `toString()` for the web environment and for others I think `.fsPath` is a stable referent.

I've added a CHANGELOG entry since in 2.9.3 we broke some user functionality and should probably do a release to ensure this does not linger.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.9.4
+
+- Fix regression in default preview styling (#501)
+
 ## 2.9.3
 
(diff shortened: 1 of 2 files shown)


Mogztter (changes requested):
As mentioned in #501 I don't mind using this workaround for now to quickly fix this bug 👍🏻 
Could you

Mogztter (approved):
Perfect, thanks 👍🏻

PR #530, opened 2022-03-03
resolves #320 set CSS variables for font size, font family and line height

### LINKED ISSUE #320: Font size / preview style too large and setting doesn't work
The preview font size option doesn't appear to do anything:

image

Compared to the Markdown extension this extension has a rather large set of fonts:

*Markdown*

image

*Asciidoc*

image

On the Asciidoctor gitter channel there were some comments about this not being great.
I think we could provide a control to scale down the size of all CSS in the preview.

### LINKED ISSUE #365: Preview: Font Size and Line Height not working
This is an example issue. Use just the headings and fill out all necessary information. Screenshots & Files and Additional Context are optional.

## Description

The Preview: Font Size and Preview: Line Height are not working.

## System Information

Visual Studio Code - Insiders
- Version: 1.50.0-insider (user setup)
- Commit: 9e505675670d65138405321a60b0df4ddec28799
- Date: 2020-09-16T06:49:37.816Z
- Electron: 9.3.0
- Chrome: 83.0.4103.122
- Node.js: 12.14.1
- V8: 8.3.110.13-electron.0
- OS: Windows_NT x64 10.0.18363

## To Reproduce

Modifying the asciidoc.preview.fontSize and/or asciidoc.preview.lineHeight does not affect the preview in any way. I have tried this with both asciidoc.preview.useEditorStyle enabled/disabled and asciidoc.use_asciidoctor_js enabled/disabled.

Mogztter:
Use `--asciidoc-font-family`, `--asciidoc-font-size` and `--asciidoc-line-height` variables.

Fix #320 
Fix #365

### DIFF
--- a/media/asciidoctor-editor.css
+++ b/media/asciidoctor-editor.css
@@ -203,12 +203,11 @@ body {
   font-family: var(--asciidoc-font-family, -apple-system, BlinkMacSystemFont, "Segoe WPC", "Segoe UI", system-ui, "Ubuntu", "Droid Sans", sans-serif);
   font-size: var(--asciidoc-font-size, 14px);
   padding: 0 26px;
-  line-height: var(--asciidoc-line-height, 22px);
+  line-height: var(--asciidoc-line-height, 1.6);
   word-wrap: break-word;
   margin: 0;
   font-weight: 400;
   font-style: normal;
-  line-height: 1;
   position: relative;
   cursor: auto;
   tab-size: 4;
@@ -308,7 +307,7 @@ table.tableblock>.title,
 .ulist>.title,
 .qlist>.title,
 .hdlist>.title {
-  line-height: 1.45;
+  line-height: calc(var(--asciidoc-line-height, 1.6) * 0.9);
   /* DISABLED color: #7a2518; */
   font-weight: 400;
   margin-top: 0;
@@ -360,14 +359,13 @@ p {
   font-family: inherit;
   font-weight: 400;
   font-size: 1em;
-  line-height: 1.6;
   margin-bottom: 1.25em;
   text-rendering: optimizeLegibility
 }
 
 p aside {
   font-size: .875em;
-  line-height: 1.35;
+  line-height: calc(var(--asciidoc-line-height, 1.6) * 0.85);
   font-style: italic
 }
 
@@ -462,7 +460,7 @@ ul,
 ol,
 dl {
   font-size: 1em;
-  line-height: 1.6;
+  line-height: var(--asciidoc-line-height, 1.6);
   margin-bottom: 1.25em;
   list-style-position: outside;
   font-family: inherit
@@ -551,7 +549,7 @@ blockquote cite a:visited {
 
 blockquote,
 blockquote p {
-  line-height: 1.6;
+  line-height: var(--asciidoc-line-height, 1.6);
   /* DISABLED color: rgba(0, 0, 0, .85) */
 }
 
@@ -565,7 +563,7 @@ blockquote p {
   h4,
   h5,
   h6 {
-    line-height: 1.2
+    line-height: calc(var(--asciidoc-line-height, 1.6) * 0.75);
   }
 
   h1 {
@@ -627,7 +625,7 @@ table tbody tr td,
 table tr td,
 table tfoot tr td {
   display: table-cell;
-  line-height: 1.6
+  line-height: var(--asciidoc-line-height, 1.6);
 }
 
 h1,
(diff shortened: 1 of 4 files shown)


danyill:
I will take a look this weekend :-)

media/asciidoctor-editor.css
```diff
@@ -200,13 +200,12 @@ body {
...
+  padding: 0 26px;
+  line-height: var(--asciidoc-line-height, 22px);
+  word-wrap: break-word;
   margin: 0;
-  /* DISABLED font-family: "Noto Serif", "DejaVu Serif", serif; */
-  /* DISABLED TODO: decide whether to use serif font
-  font-family: serif; */
   font-weight: 400;
   font-style: normal;
   line-height: 1;
```
danyill, line 211: Thank you :tada: 

Font family and font size are working nicely.

Line-height is more tricky.

On inspection, I think line 211, `line-height: 1` may be overriding line 206 `line-height: var(--asciidoc-line-height, 22px);`. After removing this line I am still not entirely successful.

Sometimes this works. I am simply putting values in the preview setting on one side and waiting for the refresh on the other side. I am trying: [0.9, 1.1, 1.2, 1.5, 2, 3, 4, 5], all of which should be effective as I understand it (we are specific a multiple of the font size).

https://developer.mozilla.org/en-US/docs/Web/CSS/line-height

Looking at it more I think it may be related to many elements having overrides on line-height in the stylesheet and the examples that I'm working with.

For example:

```
p {
  font-family: inherit;
  font-weight: 400;
  font-size: 1em;
  line-height: 1.6;
  margin-bottom: 1.25em;
  text-rendering: optimizeLegibility
}
```

As I understand it, line-height is inherited but on the computed value which I  think makes it difficult to cascade a single line-height setting through a whole document.

https://tympanus.net/codrops/css_reference/line-height/

I don't think this is worth a great deal of effort...

danyill (changes requested):
Thank you for the opportunity to review this. 

I have made a few comments but I don't think having full control over line height is worth much effort as a feature and making this comprehensive could be some work...

I think we should remove the duplicate line-height definition but anything else I'm not sure about.

media/asciidoctor-editor.css
Mogztter: Nice catch, you are absolutely right.
I think the line-height should first and foremost apply on paragraphs. The Mardown extension is using a fixed value for `h1` which I think is a good, since you probably don't want to use the same value for paragraphs and headings.

Mogztter:
> I think we should remove the duplicate line-height definition but anything else I'm not sure about.

I removed the duplicate line-height and I'm now using `calc` to compute the value relative to `var(--asciidoc-line-height, 1.6)`.
For instance, `line-height: 1;` is equivalent to `line-height: calc(var(--asciidoc-line-height, 1.6) * 0.625);` when `--asciidoc-line-height` is `1.6`.

In my opinion, it gives a decent result.

danyill (changes requested):
I can't get a preview at all at present with either the default or the editor stylesheet. Some of it is due to incorrect line-height definitions but there is something else which I am struggling to find. I'll look for a little bit more (I don't have good tools or approach for this and my :eyes: fail me).

danyill:
> In my opinion, it gives a decent result.

Very nice. I like the use of `calc`.

Although I can't find it, somehow both CSS files are invalid and on my machine I can't receive a preview. We must have somehow slightly different environments that this is working for you or I am missing a commit (?).

image

I don't get any useful debug information on this error in the extension host logs or the webview developer tools console.

I see:

>Error loading webview: Error: Could not register service workers: InvalidStateError: Failed to register a ServiceWorker: The document is in an invalid state..
>onDidChangeNotification @ workbench.desktop.main.js:787

I think this means there is a CSS definition problem somewhere as nothing else has changed but I can't find it and neither can the linter or https://jigsaw.w3.org/css-validator/validator

Mogztter:
> Although I can't find it, somehow both CSS files are invalid and on my machine I can't receive a preview. We must have somehow slightly different environments that this is working for you or I am missing a commit (?).

`line-height` was missing on a few places in `asciidoctor-default.css` but `asciidoctor-editor.css` was fine and now both are working on my end.

> I don't get any useful debug information on this error in the extension host logs or the webview developer tools console.
> I think this means there is a CSS definition problem somewhere as nothing else has changed but I can't find it and neither can the linter or [jigsaw.w3.org/css-validator/validator](https://jigsaw.w3.org/css-validator/validator)

Not sure... for reference, I'm testing using the `test-workspace/full.adoc` file.

preview

danyill (approved):
>Not sure... for reference, I'm testing using the test-workspace/full.adoc file.

Somehow my `node_modules`  were polluted and a refresh fixed it.

I have tested and I think that's a great implementation :+1:

PR #534, opened 2022-03-05
extract the report errors logic into a dedicated function

Mogztter:


danyill (approved):
LGTM, thank you :clap:

PR #539, opened 2022-03-05
resolves #443 remove Asciidoctor CLI support

### LINKED ISSUE #443: Remove Asciidoctor CLI option for preview rendering
Asciidoctor CLI access requires considerable effort to support and provide feature parity compared to Asciidoctor.js. We can't as easily read information back, use extensions etc. without duplicating effort.  Inevitably, users would like feature parity between the different options.

This issue proposes that the purpose of the preview window should not be to invoke the CLI or use external tools and this feature should be removed.

This has been discussed also in #210 and briefly in #442 previously.

Mogztter:
#### What I Did

- Remove the following options:
   - `asciidoc.asciidoctor_command`: no use anymore
   - `asciidoc.saveDocbook.docbookVersion`: only DocBook 5 is supported
   - `asciidoc.use_asciidoctor_js`: this is now the only choice

- Update the README accordingly
- Fix DocBook capitalization in the documentation
- Remove `convertUsingApplication` function

resolves #443


danyill (approved):
LGTM. I can't see anything missing and functionality appears unaffected.
Let's do it. This will also help with supporting the web environment in a clean manner.

PR #542, opened 2022-03-06
use load instead of convert

Mogztter:
#### What I Did

- Add tests on `Import.Image.getCurrentImagesDir`
- Change the signature of `getCurrentImagesDir` to make it easier to test
- Fix the regular expression that checks if the string starts with a protocol: `/^(?:[a-z]+:)?\/\//i` (remote)
- Use the `load` function in `attributeReferenceProvider.ts`, `tableOfContentsProvider.ts` and `image-paste.ts` since we only need an Asciidoctor.Document (instead of the parse/convert)

In a following pull request, I will probably remove a layer of indirection since we don't really need both `AsciidocEngine` and `AsciidocParser`.
In other words, `AsciidocEngine` has little to no value.

### DIFF
--- a/src/asciidocEngine.ts
+++ b/src/asciidocEngine.ts
@@ -64,9 +64,9 @@ export class AsciidocEngine {
     return { output, document }
   }
 
-  public async load (documentUri: vscode.Uri, source: string): Promise<any> {
+  public async load (documentUri: vscode.Uri): Promise<Asciidoctor.Document> {
     const textDocument = await vscode.workspace.openTextDocument(documentUri)
-    const { document } = await this.getEngine().parseText(source, textDocument)
+    const { document } = this.getEngine().load(textDocument)
     return document
   }
 }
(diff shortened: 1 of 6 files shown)


danyill:
> Add tests on Import.Image.getCurrentImagesDir

I'm so happy to see some tests! :grinning: 

> In a following pull request, I will probably remove a layer of indirection since we don't really need both AsciidocEngine and AsciidocParser. In other words, AsciidocEngine has little to no value.

One of the problems with the existing approach is sometimes we've been directly calling the convertUsingJavascript and there are times we've been doing it through the AsciidocEngine which is instantiated on startup.
They're not the same and there are undesirable side effects on error messages and other things so I think this would be an improvement.

danyill:
> Add tests on Import.Image.getCurrentImagesDir

I'm so happy to see some tests! :grinning: 

> In a following pull request, I will probably remove a layer of indirection since we don't really need both AsciidocEngine and AsciidocParser. In other words, AsciidocEngine has little to no value.

One of the problems with the existing approach is sometimes we've been directly calling the convertUsingJavascript and there are times we've been doing it through the AsciidocEngine which is instantiated on startup.
They're not the same and there are undesirable side effects on error messages and other things so I think this would be an improvement.

PR #547, opened 2022-03-11
move previewContentProvider in the asciidoctorWebViewConverter

Mogztter:
This is a work in progress.
The main goal is to create a suitable HTML document for VS code. It will give us more control over the generated HTML. For instance, we can use a more fitting strategy to resolve the custom stylesheet path.


src/asciidoctorWebViewConverter.ts
```diff
@@ -40,6 +121,64 @@ export class AsciidoctorWebViewConverter {
...
+            self_closing_tag_slash: '/',
+          }
+          syntaxHighlighter.$docinfo('head', node, Opal.hash2(Object.keys(opts), opts))
+        }
+      }
+      // QUESTION: should we embed font awesome?
+      // QUESTION: should we embed the following fonts (when using the default stylesheet)?
+      // Open+Sans:300,300italic,400,400italic,600,600italic
+      // Noto+Serif:400,400italic,700,700italic
+      // Droid+Sans+Mono:400,700
```
danyill, line 151: Yes, I think so. I guess this extension should be functional without internet access.

src/asciidoctorWebViewConverter.ts
```diff
@@ -49,4 +188,190 @@ export class AsciidoctorWebViewConverter {
...
+    ].join(' ')
+  }
+
+  private extensionResourcePath (mediaFile: string): string {
+    const webviewResource = this.editor.webview.asWebviewUri(vscode.Uri.joinPath(this.context.extensionUri, 'dist', mediaFile))
+    return webviewResource.toString()
+  }
+
+  private getStyles (resourceProvider: WebviewResourceProvider, resource: vscode.Uri, config: AsciidocPreviewConfiguration, state?: any): string {
+    // QUESTION: should we support `stylesdir` and `stylesheet` attributes?
```
danyill, line 317: I think a lot of users want to be able to add custom CSS have a preview that supports this. I at least often want to add a small amount of custom styling and I often have a base folder with my own Asciidoctor stylesheet (since there's no facility to incrementally add styles in Asciidoctor) which sets the theme and is something I used in many repos/documents.

src/asciidocParser.ts
```diff
@@ -125,43 +134,7 @@ export class AsciidocParser {
...
-      }
-    } else {
-      attributes = {
-        stylesdir: this.stylesdir,
-        stylesheet: 'asciidoctor-default.css@',
-      }
-    }
+    const attributes = {}
 
     // TODO: Check -- Not clear that this code is functional
```
danyill, line 139: The function works for me if I set in `settings.json`:

    "asciidoc.preview.attributes": {
        "test": "bla222h",
    }

src/features/previewConfig.ts
```diff
@@ -47,7 +49,9 @@ export class AsciidocPreviewConfiguration {
     this.fontSize = Math.max(8, +asciidocConfig.get<number>('preview.fontSize', NaN))
     this.lineHeight = Math.max(0.6, +asciidocConfig.get<number>('preview.lineHeight', NaN))
 
-    this.styles = asciidocConfig.get<string[]>('styles', [])
+    this.styles = asciidocConfig.get<string[]>('styles', []) // REMIND: unused, we should either use it or remove it!
```
danyill, line 52: I think we should remove it. Asciidoctor provides in-document mechanisms for this and I'm not sure there is a "global need for a different style" via a setting, particularly if it is non-functional now.

danyill (approved):
This is great work, you are a :rocket: 
I have made a few comments on some of the questions.

src/asciidoctorWebViewConverter.ts
Mogztter: Now that we do not rely exclusively on the `stylesheet` and `stylesdir` attributes, we can actually provide a mechanism or an option to add a stylesheet in addition to the built-in styles.

src/asciidoctorWebViewConverter.ts
Mogztter: I've added Noto Serif, Open Sans and Roboto mono

src/asciidocParser.ts
Mogztter: I will remove the comment then 👍🏻

src/features/previewConfig.ts
Mogztter: It might be a failed attempt to replicate what the Markdown extension is doing: https://github.com/microsoft/vscode/blob/3e5c7e2c570a729e664253baceaf443b69e82da6/extensions/markdown-language-features/src/features/previewConfig.ts#L49

But the Markdown extension do have a configuration named `styles`: https://github.com/microsoft/vscode/blob/3e5c7e2c570a729e664253baceaf443b69e82da6/extensions/markdown-language-features/package.json#L275-L283

> Asciidoctor provides in-document mechanisms for this and I'm not sure there is a "global need for a different style" via a setting, particularly if it is non-functional now.

I don't think it's a "global style", I think it was just a hastily copied and pasted. The Markdown extension supports a list of stylesheets and eventually I think we should too. But one thing at a time 😉

danyill:
LGTM, I tested a little. I don't really understand the changes in `src/util/dispose.ts`, can you give me a one liner?

Mogztter:
> I don't really understand the changes in src/util/dispose.ts, can you give me a one liner?

I took the code from the Markdown extension. I believe we are using an outdated version of the code. I'm slowly trying to catch up/align with the latest version of the Markdown extension.
Disposable components will extend the abstract class `Disposable` (to hide all the `_isDisposed` and `_disposables` logic)

Mogztter:
@danyill not sure if we should commit MathJax files or ignore them (i.e., `.gitignore`) and copy/include them at build time?
We could do the same for Highlight.js and FontAwesome? 

The downside is that it's harder to track changes. On the bright side, less files in the git history.

Mogztter:
I also removed fonts from Git, they are now copied at build time from npm packages: 

- https://www.npmjs.com/package/@fontsource/noto-serif
- https://www.npmjs.com/package/@fontsource/open-sans
- https://www.npmjs.com/package/@fontsource/roboto-mono

We might want to add an option to skip the `copy-assets` task when running the`build` task to save some time as we only need to run it once (except if we update the `package.json` file)

PR #558, opened 2022-03-26
Fix format of paste image default filename

Zhou-Yicheng:
There are two colons in ISO date format, only the first one was replaced, causing a %3A in the URI name.

### DIFF
--- a/src/image-paste.ts
+++ b/src/image-paste.ts
@@ -150,8 +150,7 @@ export namespace Import {
 
       const currentDateString = new Date()
         .toISOString()
-        .replace(':', '-')
-        .replace('.', '-')
+        .replace(/[:.]/g, '-')
       // default filename
       let filename = `${currentDateString}.png`
       let alttext = '' //todo:...


src/image-paste.ts
```diff
@@ -151,6 +151,7 @@ export namespace Import {
       const currentDateString = new Date()
         .toISOString()
         .replace(':', '-')
+        .replace(':', '-')
```
Mogztter, line 154: It's a shame that `replaceAll` is only available on Node 15 and greater: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll

As far as I know, the latest version of VS code is still based on Node 14.

I think it would be better to use a regular expression: 

```js
new Date().toISOString().replace(/[:.]/g, '-')
```

Mogztter:
Nice catch!
I left a suggestion, let me know what you think.

src/image-paste.ts
Zhou-Yicheng: yes, a regexp would be better

Mogztter:
All good, thanks again 👍🏻

PR #559, opened 2022-03-27
Add code folding on open block

marieflorescontact:


src/features/foldingProvider.ts
```diff
@@ -48,6 +51,34 @@ export default class AsciidocFoldingRangeProvider implements vscode.FoldingRange
...
+  private static getOpenBlockFoldingRanges (document: vscode.TextDocument) {
+    const listOfRanges = []
+    const openBlockIndexes = []
+    let end: number | undefined
+    for (let lineIndex = 0; lineIndex < document.lineCount; lineIndex++) {
+      const line = document.lineAt(lineIndex)
+      if (line.text === '--') {
+        openBlockIndexes.push(lineIndex)
+      }
+      if (line.text === '[source]') {
```
Mogztter, line 63: The documentation shows two examples but here's a valid list of open blocks:

```
[.foo.bar]
--
text
--

[f]
--
text
--

[]
--
text
--

[#bar]
--
text
--

[opts=a]
--
text
--

[foo=bar]
--
text
--
```

As you can see, the only requirement is that it must start with `[` and it must end with `]` .

src/features/foldingProvider.ts
```diff
@@ -48,6 +51,34 @@ export default class AsciidocFoldingRangeProvider implements vscode.FoldingRange
...
+        openBlockIndexes.push(lineIndex)
+        lineIndex++
+      }
+      if (line.text === '[sidebar]') {
```
Mogztter, line 67: ditto ☝🏻

src/features/foldingProvider.ts
```diff
@@ -48,6 +51,34 @@ export default class AsciidocFoldingRangeProvider implements vscode.FoldingRange
...
+        openBlockIndexes.push(lineIndex, lineIndex + 2)
```
Mogztter, line 68: I guess you are using `+ 2` to offset the block title? Any open block can have a title so something must be wrong since we don't do it for `[source]` 🤔

src/features/foldingProvider.ts
```diff
@@ -48,6 +51,34 @@ export default class AsciidocFoldingRangeProvider implements vscode.FoldingRange
...
+      }
+    }
+    for (let index = 0; index < openBlockIndexes.length; index++) {
```
Mogztter, line 71: second loop 😬

src/test/foldingProvider.test.ts
```diff
@@ -140,6 +140,92 @@ endif::[]`)
...
+--
+This is aside text.
+
+It is used to present information related to the main content.
+--
+
+this is a paragraph`)
+  assert.strictEqual(folds.length, 2, 'expecting 2 folds')
+  assert.deepStrictEqual(folds, [
+    new vscode.FoldingRange(2, 4, vscode.FoldingRangeKind.Region),
```
Mogztter, line 189: it should be:

```suggestion
    new vscode.FoldingRange(2, 8, vscode.FoldingRangeKind.Region),
```

src/test/foldingProvider.test.ts
```diff
@@ -140,6 +140,92 @@ endif::[]`)
...
+
+[source]
+--
+puts "I'm a source block!"
+--
+
+this is a paragraph`)
+  assert.strictEqual(folds.length, 1, 'expecting 1 fold')
+  assert.deepStrictEqual(folds, [
+    new vscode.FoldingRange(2, 5, vscode.FoldingRangeKind.Region),
```
Mogztter, line 206: it should be:

```suggestion
    new vscode.FoldingRange(2, 5, vscode.FoldingRangeKind.Region),
    new vscode.FoldingRange(3, 5, vscode.FoldingRangeKind.Region),
```

src/test/foldingProvider.test.ts
```diff
@@ -140,6 +140,92 @@ endif::[]`)
...
+})
+
+test('Nested open blocks  should behave like 2 separate blocks', () => {
+  const folds = getFoldsForDocument(
+    `this is a paragraph
+
+[source]
+--
+--
+puts "I'm a nested block!"
```
Mogztter, line 217: it does not produce a nested block (this is just text but it's misleading)

src/test/foldingProvider.test.ts
```diff
@@ -140,6 +140,92 @@ endif::[]`)
...
+--
+--
+
+this is a paragraph`)
+  assert.strictEqual(folds.length, 2, 'expecting 2 folds')
+  assert.deepStrictEqual(folds, [
```
Mogztter, line 223: I don't think this is a good test as the expected result is unclear.
In this case, Asciidoctor will produce the following blocks:

- An empty listing/source block
- A paragraph that contains "puts "I'm a nested block!"
- An empty open block

src/test/foldingProvider.test.ts
```diff
@@ -140,6 +140,92 @@ endif::[]`)
...
+    new vscode.FoldingRange(2, 4, vscode.FoldingRangeKind.Region),
+    new vscode.FoldingRange(6, 7, vscode.FoldingRangeKind.Region),
+  ])
+})
+
```
Mogztter, line 228: you should add additional tests:

#### given 

```
before

--
---
inside
--

after
```

#### assert

- start: 2, end: 5

#### given 

```
before

[foo=bar]
.Title
--
text
--

after
```

#### assert

- start: 2, end: 6 
- start: 4, end: 6

Should we also add a fold that starts at 3 and ends at 6?


#### given 

```
before

.Title
[foo=bar]
--
text
--

after
```

#### assert

- start: 2, end: 6 
- start: 3, end: 6 
- start: 4, end: 6

Mogztter (changes requested):
I left a few comments.
I don't think that the assertions are correct and I would like to add more test cases to make sure that the implementation is working as expected.

src/test/foldingProvider.test.ts
Mogztter: I take it back, the result is not unclear, the parsing rules are:

- An empty open block masquerading as a source block:
```
[source]
--
--
```

- A paragraph that contains "puts "I'm a nested block!"
```
puts "I'm a nested block!"
```
- An empty open block

```
--
--
```

- A paragraph

```

this is a paragraph
```

src/test/foldingProvider.test.ts
```diff
@@ -119,25 +121,158 @@ endif::[]
...
+    test('Should fold open block from dashes to the end of the document if no end block dashes', () => {
+      const folds = getFoldsForDocument(
+        `this is a paragraph
+
+--
+An open block can be an anonymous container,
+or it can masquerade as any other block.
+
+this is a paragraph`)
```
Mogztter, line 171:
```suggestion
An unterminated open block.
Fold will end at the end of the document.

this is a paragraph`)
```

Mogztter (approved):
All good, thanks 👍🏻

Issue #560, opened 2022-04-04
Add code folding on comment blocks


ggrossetie:
It would be nice to add code folding on comment blocks.

```
////
A comment block.

Notice it's a delimited block.
////
```

We should also do code folding on a group of single line comments:

```
// A single-line comment.
// Another single-line comment.
```

If single line comments are not contiguous, we should not add a code fold, for instance:

```
// A single-line comment.

// Another single-line comment.

// This is a comment.

This is a paragraph.

// This is another comment.
```

Since we already have code folding on open blocks, the following syntax is already supported:

```
[comment]
--
A comment block.

Notice it's a delimited block.
--
```

AsciiDoc also supports a comment block that can consists of a single paragraph:

```
[comment]
A paragraph comment.
Like all paragraphs, the lines must be contiguous.

Not a comment.
```

Since this syntax is harder to parse, we won't support it right away.

Documentation: https://docs.asciidoctor.org/asciidoc/latest/comments/

marieflorescontact:
:wave:
I can submit a pull request for this issue!

PR #565, opened 2022-04-09
Add code folding on single line comment

marieflorescontact:
2nd step of #560 issue: code folding on single line comment.

### DIFF
--- a/src/features/foldingProvider.ts
+++ b/src/features/foldingProvider.ts
@@ -99,16 +99,48 @@ export default class AsciidocFoldingRangeProvider implements vscode.FoldingRange
     }
   }
 
+  private static handleSingleLineCommentFoldingRanges (singleLineCommentStartIndexes: any[], foldingRanges: any[], lineIndex: number, lineText: string,
+    documentLineCount: number) {
+    if (lineText.startsWith('//')) {
+      if (singleLineCommentStartIndexes.length === 0 && lineIndex < documentLineCount - 1) {
+        singleLineCommentStartIndexes.push(lineIndex)
+      } else {
+        // comment on last line of the document
+        const startIndex = singleLineCommentStartIndexes.pop()
+        if (lineIndex > startIndex) {
+          foldingRanges.push(new vscode.FoldingRange(
+            startIndex,
+            lineIndex,
+            FoldingRangeKind.Region)
+          )
+        }
+      }
+    } else {
+      if (singleLineCommentStartIndexes.length !== 0) {
+        const startIndex = singleLineCommentStartIndexes.pop()
+        const endIndex = lineIndex - 1
+        if (endIndex > startIndex) {
+          foldingRanges.push(new vscode.FoldingRange(
+            startIndex,
+            endIndex,
+            FoldingRangeKind.Region))
+        }
+      }
+    }
+  }
+
   private static getBlockFoldingRanges (document: vscode.TextDocument) {
     const foldingRanges = []
     const openBlockIndexes = []
     const commentBlockIndexes = []
+    const singleLineCommentStartIndexes = []
     const documentLineCount = document.lineCount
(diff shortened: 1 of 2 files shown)


src/features/foldingProvider.ts
```diff
@@ -99,16 +99,43 @@ export default class AsciidocFoldingRangeProvider implements vscode.FoldingRange
     }
   }
 
+  private static handleSingleLineCommentFoldingRanges (singleLineCommentStartIndexes: any[], foldingRanges: any[], lineIndex: number, lineText: string,
+    isPreviousLineSingleCommented: boolean) {
```
Mogztter, line 103: I don't think we need the `isPreviousLineSingleCommented` variable.
If the line starts with `//`, there are 2 cases:

- it's the beginning of one or more single line comments
- it's the continuation of single line comments

If the line does not start with `//`, there are 3 cases:

- it's the end of only one single line comment
- it's the end of two or more single line comments
- it's neither

**Logic**
- When the line starts with `//` and `singleLineCommentStartIndexes` is empty, we should add the current line index to it: `singleLineCommentStartIndexes.push(lineIndex)`.
- When the line starts with `//` and `singleLineCommentStartIndexes` is not empty, it's a continuation of single line comments (and we don't need to do anything).
- When the line does not start with `//` and `singleLineCommentStartIndexes` is empty, we don't need to anything.
- When the line does not start with `//` and `singleLineCommentStartIndexes` is not empty, we should check if the current line number is greater than `singleLineCommentStartIndexes.pop() + 1`. If that's the case, then we have two or more single line comments. If not, then it's a single line comment.

**Pseudo-code**
```js
if (lineText.startsWith('//')) {
  if (singleLineCommentStartIndexes.length === 0) {
    singleLineCommentStartIndexes.push(lineIndex)
  }
} else {
  if (singleLineCommentStartIndexes.length !== 0) {
    const startIndex = singleLineCommentStartIndexes.pop()
    const endIndex = lineIndex - 1
    if (endIndex > startIndex) {
      foldingRanges.push(new vscode.FoldingRange(
          startIndex,
          endIndex,
          FoldingRangeKind.Region))
      }
    }
  }
}
```

src/features/foldingProvider.ts
```diff
@@ -99,16 +99,43 @@ export default class AsciidocFoldingRangeProvider implements vscode.FoldingRange
...
+    }
+    if (lineText.slice(0, 2) === '//' && !isPreviousLineSingleCommented) {
+      singleLineCommentStartIndexes.push(lineIndex)
+      isPreviousLineSingleCommented = true
+    }
+    if (lineText.slice(0, 2) !== '//' && isPreviousLineSingleCommented) {
+      const startIndex = singleLineCommentStartIndexes.pop()
+      const endIndex = lineIndex - 1
+      if (endIndex - startIndex > 0) {
+        foldingRanges.push(new vscode.FoldingRange(
```
Mogztter, line 118: We should also create a folding range when `singleLineCommentStartIndexes` is not empty and we are on the last line. For instance:

```
This is a paragraph.

// This is a comment.
// The last line of the document is also a comment!
```

src/features/foldingProvider.ts
```diff
@@ -99,16 +99,50 @@ export default class AsciidocFoldingRangeProvider implements vscode.FoldingRange
     }
   }
 
+  private static handleSingleLineCommentFoldingRanges (singleLineCommentStartIndexes: any[], foldingRanges: any[], lineIndex: number, lineText: string,
+    documentLineCount: number) {
+    if (lineText.slice(0, 2) === '//') {
```
Mogztter, line 104: We can use `startsWith`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith

Mogztter (approved):
All good, thanks 👍🏻

Issue #422, opened 2021-07-19
How to register asciidoctor.js extensions?


chtenb:
We can set attributes using the `asciidoc.preview.attributes` option, like `    'source-highlighter': 'highlightjs-ext',`.
However, this particular highlighter is an extension which is normally registered like `highlightJsExt.register(asciidoctor.Extensions)` in a asciidoctor.js script. How do I make sure that this plugin is registered in asciidoctor-vscode?

chtenb:
Personally I used it on a per-workspace basis. I have a project which uses asciidoctor extensively and uses the following script to produce html. In my usecase it would be nice if the VSCode preview would match the output of this script, because right now the VSCode preview doesn't work well enough for me. The difference is too big w.r.t. my eventual html output.

> For an initial implementation, would it be sufficient to have a "Use extensions" option and then perhaps a field for comma delimited references to absolute paths to modules which export a register function (e.g. like asciidoctor-chart).

That sounds okay.

Would it be hard to hook a script like below into the extension? That way everyone can just do whatever they want.
Maybe it would need to have a standardized name, like `asciidoctor.rc.js` or something.

```js
const Asciidoctor = require('asciidoctor')
const kroki = require('asciidoctor-kroki')
const highlightJsExt = require('asciidoctor-highlight.js')
const glob = require('glob')

var asciidoctor = Asciidoctor()

kroki.register(asciidoctor.Extensions)
highlightJsExt.register(asciidoctor.Extensions)

var options = {
  'safe': 'unsafe',
  'standalone': true,
  'attributes': {
    'linkcss': true,
    'source-highlighter': 'highlightjs-ext',
    'docinfodir': '../res/',
    'docinfo': 'shared',
    'stylesheet': '../res/asciidoc.css',
    'kroki-default-options': 'inline',
    'allow-uri-read': true
  }
}

if (process.argv[2]) {

  options['to_file'] = false
  var html = asciidoctor.convertFile(process.argv[2], options)
  console.log(html)

} else {

  glob('blog/*.adoc', function (err, files) {
    if (err) {
      console.log(err)
    } else {
      files.forEach(function (file) {
        asciidoctor.convertFile(file, options)
      });
    }
  });

}
```

PR #567, opened 2022-04-24
Add emoji support

YoshihideShirai:
Add emoji support.
It uses "Asciidoctor Emoji Extension".
https://github.com/Mogztter/asciidoctor-emoji

"asciidoc.use_emoji" setting is added. It set at false by default.


Mogztter:
Thanks for your contribution.
The Asciidoctor Emoji is not an "official" extension (not part of the `@asciidoctor` organization) and is not very popular. I don't think we should include it by default.

Having said that, we should find a way to register "custom" extensions, see: https://github.com/asciidoctor/asciidoctor-vscode/issues/422

danyill:
I agree. 

I think that once #569 is merged that allowing the capability for users to provide their own extensions prevents us from having to ship many extensions and we should close this.

Mogztter:
In the next version it will be possible to register custom extensions thanks to @YoshihideShirai 🥇

PR #573, opened 2022-06-13
Fixes for opening links to local files

marieflorescontact:


src/features/preview.ts
```diff
@@ -433,25 +433,33 @@ export class AsciidocPreview {
     vscode.workspace.openTextDocument(this._resource).then(vscode.window.showTextDocument)
   }
 
-  private async onDidClickPreviewLink (href: string) {
-    let [hrefPath, fragment] = decodeURIComponent(href).split('#')
-    // From Markdown plugin
-    if (hrefPath[0] !== '/') {
+  private resolveDocumentLink (href: string): { hrefPath: string, fragment: string } {
+    let [hrefPath, fragment] = href.split('#').map((c) => decodeURIComponent(c))
+    if (hrefPath.startsWith('file://')) {
```
Mogztter, line 438: This is not necessarily an absolute path. For instance:

- `file:///path/to/dir` is an absolute path
- `file://../../path/to/dir` is a relative path

Both paths start with `file://` but only the first one is an absolute path. I think we should remove the protocol `file://`.

src/features/preview.ts
```diff
@@ -433,25 +433,33 @@ export class AsciidocPreview {
...
+      hrefPath = vscode.Uri.parse(hrefPath).fsPath
+    }
+    if (!hrefPath.startsWith('/')) {
+      // Relative path. Resolve relative to the file
       hrefPath = path.join(path.dirname(this.resource.fsPath), hrefPath)
-    } else {
-      // Handle any normalized file paths
-      hrefPath = vscode.Uri.parse(hrefPath.replace('/file', '')).fsPath
     }
+    return { hrefPath: hrefPath, fragment: fragment }
```
Mogztter, line 446: We could return `{ path: hrefPath, fragment: fragment }`. That way, we don't need to destructure the object when calling `executeCommand` we can directly pass the object:

```ts
vscode.commands.executeCommand('_asciidoc.openDocumentLink', targetResource)
```

Mogztter (approved):
Thanks 👍🏻

PR #578, opened 2022-06-22
Prevent code execution doesn't work

YoshihideShirai:
Hi, @Mogztter .

I found a bug in #569 implement.
I create fix MR.
Can you check it?

### DIFF
--- a/src/security.ts
+++ b/src/security.ts
@@ -192,9 +192,10 @@ export class AsciidoctorExtensionsSecurityPolicyArbiter {
   }
 
   public async confirmAsciidoctorExtensionsTrustMode (extensionsCount: number): Promise<boolean> {
-    if (this.asciidoctorExtensionsAuthorsTrusted() !== undefined) {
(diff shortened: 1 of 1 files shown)


Mogztter:
I've updated the logic a bit so I don't know if this is an actual bug or the expected behavior. Could you please describe what you did, what was your expectation and what was the actual behavior?

Once you gave your trust the extension won't ask again *unless* you disable (and enable again) extensions registration from the settings.
It's also possible to use the command "Manage Asciidoctor.js Extensions Trust Mode" to revoke your trust (denied).

YoshihideShirai:
Sorry for my poor explanation.

Trust mode is always allow after users select trust mode.
So, it is allow despite of selecting deny.
I think there is some confusion in the definition of the return value of conformAscidoctorExtensionsTrustMode().

Mogztter:
Oh I see, it makes sense now.
I think you are right, let me test again.

Mogztter:
You are indeed correct, nice catch!

PR #580, opened 2022-06-23
activate Antora support

marieflorescontact:


package.json
```diff
@@ -427,6 +427,16 @@
 					"type": "boolean",
 					"default": false,
 					"description": "%asciidoc.useWorkspaceRoot.desc%"
+				},
+				"asciidoc.antora": {
```
Mogztter, line 431: I think we should create a category for Antora:

```suggestion
				"asciidoc.antora.enableAntoraSupport": {
```

package.json
```diff
@@ -427,6 +427,16 @@
...
+					"type": "boolean",
+					"default": false,
+					"description": "%asciidoc.antora.desc%"
+				},
+				"asciidoc.antoraConfigIsSet": {
```
Mogztter, line 436: As far as I know, we don't need to explicitly declare this configuration since we don't want to configure it using "User Settings".

src/extension.ts
```diff
@@ -69,6 +77,25 @@ export function activate (context: vscode.ExtensionContext) {
   commandManager.register(new commands.SaveHTML(engine))
   commandManager.register(new commands.SaveDocbook(engine))
 
+  const antora = new AntoraSupport()
+  if (await antora.isDetected()) {
+    const config = vscode.workspace.getConfiguration('asciidoc', null)
+    if (config.get('antora') === false && config.get('antoraConfigIsSet') === false) {
```
Mogztter, line 83: We should use https://code.visualstudio.com/api/references/vscode-api#WorkspaceConfiguration to check that `isEnableAntoraSupportSettingDefined` is `undefined`.

src/util/antoraSupport.ts
```diff
@@ -0,0 +1,34 @@
...
+    const startPath = vscode.window.activeTextEditor.document.uri.fsPath
+    // check that we stay in the current workspace
+    const currentWorkspaceFolderName = this.findCurrentWorkspaceFolder().name
+    if (startPath.indexOf(currentWorkspaceFolderName) < startPath.indexOf('modules')) {
+      return startPath.substring(0, startPath.indexOf('modules'))
+    }
+    return undefined
+  }
+
+  public async getAttributes (): Promise<any| undefined> {
```
Mogztter, line 25: Type is probably a list of key/value, we shouldn't use `any`

src/util/antoraSupport.ts
```diff
@@ -0,0 +1,34 @@
...
+    try {
+      const doc = yaml.load(fs.readFileSync(this.findModulesDirPath() + 'antora.yml', 'utf8'))
+      return doc.asciidoc.attributes
+    } catch (err) {
+      await vscode.window.showErrorMessage('Error parsing antora.yml, ' + err.toString())
+      return undefined
```
Mogztter, line 31: if we return `undefined`, then we will call `Object.entries(undefined)` in `src/features/antoraCompletionProvider.ts` (which will fail)

src/util/antoraSupport.ts
```diff
@@ -0,0 +1,34 @@
...
+    const antoraYmlFiles = await vscode.workspace.findFiles('**/antora.yml', '/node_modules/', 10)
+    return antoraYmlFiles.length > 0
+  }
+
+  public findCurrentWorkspaceFolder (): WorkspaceFolder {
+    return vscode.workspace.getWorkspaceFolder(vscode.Uri.file(vscode.window.activeTextEditor.document.uri.fsPath))
+  }
+
+  public findModulesDirPath (): string | undefined {
+    const startPath = vscode.window.activeTextEditor.document.uri.fsPath
```
Mogztter, line 16: we should use the TextDocument.uri (https://code.visualstudio.com/api/references/vscode-api#Uri) from the `provideCompletionItems` method

src/util/antoraSupport.ts
```diff
@@ -0,0 +1,34 @@
...
+    // check that we stay in the current workspace
+    const currentWorkspaceFolderName = this.findCurrentWorkspaceFolder().name
+    if (startPath.indexOf(currentWorkspaceFolderName) < startPath.indexOf('modules')) {
```
Mogztter, line 19: That's not reliable, what if we have a workspace named `foo` and the following path: `/path/to/foo/docs/modules/ROOT/pages/foo/page.adoc`?

package.json
```diff
@@ -427,6 +427,11 @@
 					"type": "boolean",
 					"default": false,
 					"description": "%asciidoc.useWorkspaceRoot.desc%"
+				},
+				"asciidoc.antora.enableAntoraSupport": {
+					"type": "boolean",
+					"default": false,
+					"description": "%asciidoc.antora.desc%"
```
Mogztter, line 434: you need to add a translation in `package.nls.json` to explain what this setting does

src/extension.ts
```diff
@@ -69,6 +77,29 @@ export function activate (context: vscode.ExtensionContext) {
   commandManager.register(new commands.SaveHTML(engine))
   commandManager.register(new commands.SaveDocbook(engine))
 
+  const antora = new AntoraSupport()
+  if (await antora.isDetected()) {
+    const userSettings = vscode.workspace.getConfiguration('asciidoc', null)
+    await userSettings.get('antora.enableAntoraSupport')
```
Mogztter, line 83:
```suggestion
```

src/extension.ts
```diff
@@ -69,6 +77,29 @@ export function activate (context: vscode.ExtensionContext) {
...
+        context.subscriptions.push(vscode.languages.registerCompletionItemProvider(selector, new AntoraCompletionProvider(), '{'))
+        await userSettings.update('antora.enableAntoraSupport', true)
+        await workspaceSettings.update('antoraSupportSetting', true)
+      }
+      if (answer === 'no thanks') {
+        await workspaceSettings.update('antoraSupportSetting', true)
+      }
+      if (answer === undefined) {
+        // do nothing
+      }
```
Mogztter, line 99:
```suggestion
```

src/extension.ts
```diff
@@ -69,6 +77,29 @@ export function activate (context: vscode.ExtensionContext) {
...
+    if (userSettings.get('antora.enableAntoraSupport') === false && isEnableAntoraSupportSettingDefined === undefined) {
+      const answer = await vscode.window.showInformationMessage('We detect that you are working with Antora. Do you want to active Antora support?', 'yes', 'no thanks')
+      if (answer === 'yes') {
+        context.subscriptions.push(vscode.languages.registerCompletionItemProvider(selector, new AntoraCompletionProvider(), '{'))
+        await userSettings.update('antora.enableAntoraSupport', true)
+        await workspaceSettings.update('antoraSupportSetting', true)
+      }
+      if (answer === 'no thanks') {
+        await workspaceSettings.update('antoraSupportSetting', true)
+      }
```
Mogztter, line 96: we can do something like:

```ts
if (answer !== undefined) {
  await workspaceSettings.update('antoraSupportSetting', true)
  const enableAntoraSupport = answer === 'yes' ? true : (answer === 'no thanks' ? false : undefined)
  await userSettings.update('antora.enableAntoraSupport', enableAntoraSupport)
  if (enableAntoraSupport) {
    context.subscriptions.push(vscode.languages.registerCompletionItemProvider(selector, new AntoraCompletionProvider(), '{'))
  }
}
```

package.json
Mogztter, line 434: you can use `markdownDescription` instead of `description` to use Markdown syntax in the description

src/util/antoraSupport.ts
```diff
@@ -0,0 +1,36 @@
...
+    return antoraYmlFiles.length > 0
+  }
+
+  public async findModulesDirPath (textDocument: vscode.TextDocument): Promise<any> {
+    const pathToAsciidocFile = textDocument.uri.fsPath
+    const antoraYmlFiles = await vscode.workspace.findFiles('**/antora.yml', '/node_modules/', 100)
+    // check if antora.yml file matches with textDocument content source root
+    try {
+      for (let i = 0; i < antoraYmlFiles.length; i++) {
+        if (pathToAsciidocFile.startsWith(path.dirname(antoraYmlFiles[i].path))) {
```
Mogztter, line 18: (nitpick) you can use `for ... of` (simpler syntax):


```suggestion
      for (const antoraYmlFile of antoraYmlFiles) {
        if (pathToAsciidocFile.startsWith(path.dirname(antoraYmlFile.path))) {
```

src/util/antoraSupport.ts
```diff
@@ -0,0 +1,36 @@
...
+          return pathToAsciidocFile.substring(0, pathToAsciidocFile.indexOf('modules'))
```
Mogztter, line 19: I think we should revisit this function and return a path to an `antora.yml` file.
If no `antora.yml` is found or if none of them are applicable then we could return `undefined` or throw an `Error`.

src/util/antoraSupport.ts
```diff
@@ -0,0 +1,36 @@
...
+        }
+      }
+    } catch (err) {
+      await vscode.window.showErrorMessage('Antora modules is out of the current workspace, ' + err.toString())
```
Mogztter, line 23: I don't think we should show an error message. If we cannot find any `antora.yml` that's fine.
Using `console.log` is probably enough.

src/util/antoraSupport.ts
```diff
@@ -0,0 +1,36 @@
...
+    }
+  }
+
+  public async getAttributes (textDocument: vscode.TextDocument): Promise<{ [key: string]: string }> {
+    try {
+      const doc = yaml.load(fs.readFileSync(await this.findModulesDirPath(textDocument) + 'antora.yml', 'utf8'))
+      return doc.asciidoc.attributes
+    } catch (err) {
+      await vscode.window.showErrorMessage('Error parsing antora.yml, ' + err.toString())
```
Mogztter, line 32: Again, I don't think we should show an error message. It's not a big deal if the autocompletion does not show any attributes from `antora.yml` if the file is not properly formatted as YAML.

In my opinion, using `console.log` with the full path to the `antora.yml` file is enough:

```js
console.log(`Unable to parse /path/to/docs/antora.yml, cause: err.toString()`)
```

src/util/antoraSupport.ts
```diff
@@ -0,0 +1,36 @@
...
+    const antoraYmlFiles = await vscode.workspace.findFiles('**/antora.yml', '/node_modules/', 100)
+    // check if antora.yml file matches with textDocument content source root
+    try {
+      for (const antoraYmlFile of antoraYmlFiles) {
+        if (pathToAsciidocFile.startsWith(path.dirname(antoraYmlFile.path))) {
+          return antoraYmlFile.path
+        }
+      }
+    } catch (err) {
+      console.log('Antora file not applicable to current asciidoc document, ' + err.toString())
```
Mogztter, line 23: `try/catch` is superfluous since the code cannot throw an exception

src/util/antoraSupport.ts
```diff
@@ -0,0 +1,36 @@
...
+    return antoraYmlFiles.length > 0
+  }
+
+  public async getAntoraYmlPath (textDocument: vscode.TextDocument): Promise<any> {
+    const pathToAsciidocFile = textDocument.uri.fsPath
+    const antoraYmlFiles = await vscode.workspace.findFiles('**/antora.yml', '/node_modules/', 100)
+    // check if antora.yml file matches with textDocument content source root
+    try {
+      for (const antoraYmlFile of antoraYmlFiles) {
+        if (pathToAsciidocFile.startsWith(path.dirname(antoraYmlFile.path))) {
```
Mogztter, line 18: we should make sure that the `pathToAsciidocFile` starts with `/path/to/workspace/repository/modules/<module>/pages`.


```js
// pseudo-code (untested)
const asciidocFilename = path.basename(pathToAsciidocFile)
const modulesPath = path.join(path.dirname(antoraYmlFile.path), 'modules')
if (pathToAsciidocFile.startsWith(modulesPath) && pathToAsciidocFile.slice(modulesPath.length).match(/^\/[^/]+\/pages\/.*/)) {
  return antoraYmlFile.path
}
```

Mogztter: you can add the following test case:

```
/path/to/workspace/repository/antora.yml
/path/to/workspace/repository/README.adoc
```

`README.adoc` should not have Antora support enable (because it does not belong to an Antora module)

src/util/antoraSupport.ts
```diff
@@ -0,0 +1,36 @@
...
+      console.log('Antora file not applicable to current asciidoc document, ' + err.toString())
+    }
+  }
+
+  public async getAttributes (textDocument: vscode.TextDocument): Promise<{ [key: string]: string }> {
+    try {
+      const doc = yaml.load(fs.readFileSync(await this.getAntoraYmlPath(textDocument), 'utf8'))
+      return doc.asciidoc.attributes
+    } catch (err) {
+      console.log('Unable to parse /path/to/docs/antora.yml, cause:' + err.toString())
```
Mogztter, line 32: `/path/to/docs/antora.yml` should be the *actual* path to the `antora.yml` file.

src/extension.ts
```diff
@@ -69,6 +77,25 @@ export function activate (context: vscode.ExtensionContext) {
...
+      if (answer !== undefined) {
+        await workspaceState.update('antoraSupportSetting', true)
+        const enableAntoraSupport = answer === 'yes' ? true : (answer === 'no thanks' ? false : undefined)
+        await workspaceConfiguration.update('antora.enableAntoraSupport', enableAntoraSupport)
+        if (enableAntoraSupport) {
+          context.subscriptions.push(vscode.languages.registerCompletionItemProvider(selector, new AntoraCompletionProvider(), '{'))
+        }
+      }
+    }
+  }
```
Mogztter, line 97: We should move this logic when an AsciiDoc file is opened using the listener `onDidOpenTextDocument` on a `workspace`.
As a result, we can ask to enable Antora support only when we find a valid `antora.yml` for the given `TextDocument` 🎉

src/util/antoraSupport.ts
```diff
@@ -0,0 +1,39 @@
...
+      console.log('Antora file not applicable to current asciidoc document.')
+    }
+  }
+
+  public async getAttributes (textDocument: vscode.TextDocument): Promise<{ [key: string]: string }> {
+    try {
+      const doc = yaml.load(fs.readFileSync(await this.getAntoraYmlPath(textDocument), 'utf8'))
+      return doc.asciidoc.attributes
+    } catch (err) {
+      console.log(`Unable to parse ${await this.getAntoraYmlPath(textDocument)}, cause:` + err.toString())
```
Mogztter, line 31: Since we call `await this.getAntoraYmlPath(textDocument)` again we will do the process twice. We should extract the result of the method call:

```js
const antoraYmlPath = await this.getAntoraYmlPath(textDocument)
```

src/util/antoraSupport.ts
```diff
@@ -0,0 +1,36 @@
...
+    const pathToAsciidocFile = textDocument.uri.fsPath
+    const antoraYmlFiles = await vscode.workspace.findFiles('**/antora.yml', '/node_modules/', 100)
+    // check if antora.yml file matches with textDocument content source root
+    for (const antoraYmlFile of antoraYmlFiles) {
+      const asciidocFilename = path.basename(pathToAsciidocFile)
+      const modulesPath = path.join(path.dirname(antoraYmlFile.path), 'modules')
+      if (pathToAsciidocFile.startsWith(modulesPath) && pathToAsciidocFile.slice(modulesPath.length).match(/^\/[^/]+\/pages\/.*/) && asciidocFilename !== 'README.adoc') {
+        return antoraYmlFile.path
+      }
+      console.log('Antora file not applicable to current asciidoc document.')
```
Mogztter, line 17: Let's say we found 5 `antora.yml` and none are applicable, we will log:

```
Antora file not applicable to current asciidoc document.
Antora file not applicable to current asciidoc document.
Antora file not applicable to current asciidoc document.
Antora file not applicable to current asciidoc document.
Antora file not applicable to current asciidoc document.
```

The log message does not provide any context (and it says the same thing over and over). It would be better to log something like:

```
Unable to find an applicable Antora configuration file in [/path/to/antora.yml, /path/to/dir/antora.yml, /path/to/dir/sub/antora.yml] for the AsciiDoc document /path/to/page.adoc" 
```

We could also add a log when we found an applicable antora.yml file:

```
Found an Antora configuration file at /path/to/antora.yml for the AsciiDoc document /path/to/page.adoc
```

src/util/antoraSupport.ts
```diff
@@ -0,0 +1,36 @@
...
+export default class AntoraSupport {
+
+  public async getAntoraYmlPath (textDocument: vscode.TextDocument): Promise<any> {
+    const pathToAsciidocFile = textDocument.uri.fsPath
+    const antoraYmlFiles = await vscode.workspace.findFiles('**/antora.yml', '/node_modules/', 100)
+    // check if antora.yml file matches with textDocument content source root
+    for (const antoraYmlFile of antoraYmlFiles) {
+      const asciidocFilename = path.basename(pathToAsciidocFile)
+      const modulesPath = path.join(path.dirname(antoraYmlFile.path), 'modules')
+      if (pathToAsciidocFile.startsWith(modulesPath) && pathToAsciidocFile.slice(modulesPath.length).match(/^\/[^/]+\/pages\/.*/) && asciidocFilename !== 'README.adoc') {
```
Mogztter, line 14: We can remove `&& asciidocFilename !== 'README.adoc'` (it was an example)

Mogztter (changes requested):

src/extension.ts
```diff
@@ -19,8 +19,9 @@ import { AsciidocFileIncludeAutoCompletionMonitor } from './util/includeAutoComp
 import { AttributeReferenceProvider } from './features/attributeReferenceProvider'
 import { BuiltinDocumentAttributeProvider } from './features/builtinDocumentAttributeProvider'
 import AsciidocFoldingRangeProvider from './features/foldingProvider'
+import { AntoraSupportManager } from './features/antora/antoraSupport'
 
-export function activate (context: vscode.ExtensionContext) {
+export async function activate (context: vscode.ExtensionContext) {
```
Mogztter, line 24: We don't need `async` since we don't use `await`

src/features/antora/antoraSupport.ts
```diff
@@ -0,0 +1,86 @@
...
+      const workspaceConfiguration = vscode.workspace.getConfiguration('asciidoc', null)
+      // look for Antora settings in workspaceState
+      const workspaceState: vscode.Memento = this.context.workspaceState
+      const isEnableAntoraSupportSettingDefined = await workspaceState.get('antoraSupportSetting')
+      if (isEnableAntoraSupportSettingDefined === true) {
+        if (workspaceConfiguration.get('antora.enableAntoraSupport') === true) {
+          this.dispose()
+        } else {
+          // no op
+        }
```
Mogztter, line 31: If the Antora support is enabled we `dispose` something? I believe the idea was to register the completionItemProvider above only when the Antora support is enabled?

src/features/antora/antoraSupport.ts
```diff
@@ -0,0 +1,86 @@
...
+      } else {
+        if (await getValidConfig(textDocument)) {
+          const answer = await vscode.window.showInformationMessage('We detect that you are working with Antora. Do you want to active Antora support?', 'yes', 'no thanks')
+          if (answer !== undefined) {
+            await workspaceState.update('antoraSupportSetting', true)
+            const enableAntoraSupport = answer === 'yes' ? true : (answer === 'no thanks' ? false : undefined)
+            await workspaceConfiguration.update('antora.enableAntoraSupport', enableAntoraSupport)
+            if (enableAntoraSupport) {
+              this.dispose()
```
Mogztter, line 40: Here we should register the completionItemProvider if it's not already registered. If the user does not want to enable Antora support then we should not register it.

src/test/antoraCompletionProvider.test.ts
```diff
@@ -0,0 +1,20 @@
...
+suite('Antora CompletionsProvider', () => {
+  setup(() => {
+    root = vscode.workspace.workspaceFolders[0].uri.fsPath
+  })
+  test('Should return completion items', async () => {
+    const provider = new AntoraCompletionProvider()
+    const file = await vscode.workspace.openTextDocument(vscode.Uri.file(`${root}/antora/multiComponents/api/modules/auth/pages/jwt/page2.adoc`))
+    const completionsItems = await provider.provideCompletionItems(file, new Position(3, 1))
+    // dumb test, WIP
+    assert.strictEqual(completionsItems.length, 3)
```
Mogztter, line 18: we should probably assert that the completion items contain the right data

src/test/antoraSupport.test.ts
```diff
@@ -0,0 +1,60 @@
+import * as assert from 'assert'
+import 'mocha'
+import * as vscode from 'vscode'
+import { getAntoraConfig } from '../features/antora/antoraSupport'
+
+let root
+
+suite('Antora Support', () => {
```
Mogztter, line 8: That's pretty good!
I think we should add a test on `page1.adoc`.

src/test/antoraSupport.test.ts
```diff
@@ -0,0 +1,60 @@
...
+    root = vscode.workspace.workspaceFolders[0].uri.fsPath
+  })
+  test('Should return Antora config', async () => {
+    const expectedPath = `${root}/antora/multiComponents/api/antora.yml`
+    const file = await vscode.workspace.openTextDocument(vscode.Uri.file(`${root}/antora/multiComponents/api/modules/auth/pages/jwt/page2.adoc`))
+    const actualUri = await getAntoraConfig(file)
+    assert.strictEqual(actualUri.fsPath, expectedPath)
+  })
+
+  test('Should return Antora config 2', async () => {
```
Mogztter, line 19: we should find a better title 😉

src/test/antoraSupport.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  })
+
+  test('Should not return Antora config for document outside "modules" Antora folder', async () => {
+    const expected = undefined
+    const file = await vscode.workspace.openTextDocument(vscode.Uri.file(`${root}/antora/multiComponents/api/modules/writer-guide.adoc`))
+    const actualUri = await getAntoraConfig(file)
+    assert.strictEqual(actualUri, expected)
+  })
+
+  test('Should not return Antora config for document outside of workspace', async () => {
```
Mogztter, line 54: since we are doing the same thing we could introduce a function:

```ts
async function testGetAntoraConfig (asciidocPath, antoraConfigExpectedFsPath) {
    const file = await vscode.workspace.openTextDocument(vscode.Uri.file(`${root}/${asciidocPath}`))
    const antoraConfigUri = await getAntoraConfig(file)
    if (antoraConfigExpectedPath === undefined) {
        assert.strictEqual(antoraConfigUri, undefined)
    } else {
        assert.strictEqual(antoraConfigUri.fsPath, antoraConfigExpectedFsPath)
    }
}
```

Then we could call it in each test:

```ts
test('Should return Antora config', async () => testGetAntoraConfig(
  'antora/multiComponents/api/modules/auth/pages/jwt/page2.adoc',
  `${root}/antora/multiComponents/api/antora.yml`
))

test('Should return Antora config for document inside a directory wich has the same name as the workspace', async () => testGetAntoraConfig(
  'antora/multiComponents/api/modules/auth/pages/modules/multiComponents/page6.adoc',
  `${root}/antora/multiComponents/api/antora.yml`
))

test('Should not return Antora config for document outside "modules" Antora folder', async () => testGetAntoraConfig(
  'antora/multiComponents/api/modules/writer-guide.adoc',
  undefined
))
```

We can even generate the `test` block by extracting test cases data:

```ts
const testCases = [
  {
    title: 'Should return Antora config',
    asciidocPath: 'antora/multiComponents/api/modules/auth/pages/jwt/page2.adoc',
    antoraConfigExpectedFsPath: `${root}/antora/multiComponents/api/antora.yml`
  }
  // ...
]
for (const testCase of testCases) {
  test(testCase.title, async () => testGetAntoraConfig(testCase.asciidocPath, testCase.antoraConfigExpectedFsPath))
}
```

package.json
```diff
@@ -607,5 +613,11 @@
 		"asciidoctor-kroki": "0.15.4",
 		"uuid": "8.3.2",
 		"vscode-nls": "5.0.0"
+	},
+	"__metadata": {
+		"id": "c1309cc2-f420-46a3-b2be-ca04f4d9e51b",
+		"publisherDisplayName": "asciidoctor",
+		"publisherId": "b139f5ca-8a7c-4029-b14a-59f66583c82c",
+		"isPreReleaseVersion": false
```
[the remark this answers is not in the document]
marieflorescontact: nop...I was thinking it came from the changes i rebased from origin/main...
I've found this [issue](https://github.com/microsoft/vscode/issues/148975) , it seems to be the same problem

src/features/antora/antoraSupport.ts
marieflorescontact: yes you 're right, I've mixed everything up by changing the logic

src/test/antoraCompletionProvider.test.ts
marieflorescontact: yes...but for the moment I may have an issue with the insert text because all the assertions failed :(
```
 	  -    "insertText": "{xrefstyle,short@}"
      +    "insertText": "xrefstyle,short@"
```

package.json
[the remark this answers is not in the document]
Mogztter: Apparently this is expected but it's not clear if this is safe to keep them or not 🤔 
I would say remove them for now.

src/test/antoraCompletionProvider.test.ts
Mogztter: yes the expected value of `insertText` should be `{xrefstyle}`. The problem is that you are using:

```js
return Object.entries(attributes).map((key) => {
  // ...
})
```

`entries` will produce a list of `[key, value]` and you are using `let insertText = key.toString()` so you will get the stringify value of an array that contains the key and value:

```js
['xrefstyle', 'short@'].toString() // xrefstyle,short@
```

Instead, you should use the following:

```js
return Object.entries(attributes).map(([key, value]) => {
  const completionItem = new vscode.CompletionItem({
      label: key,
      description: value,
    }, vscode.CompletionItemKind.Text)
    let insertText = key
    insertText = prefix !== '{' ? `{${insertText}` : insertText
    insertText = suffix !== '}' ? `${insertText}}` : insertText
    completionItem.insertText = insertText
    return completionItem
})
```

src/features/antora/antoraCompletionProvider.ts
```diff
@@ -0,0 +1,22 @@
...
+      const completionItem = new vscode.CompletionItem({
+        label: key[0].toString(),
+        description: key[1].toString(),
+      }, vscode.CompletionItemKind.Text)
+      let insertText = key[0].toString()
+      insertText = prefix !== '{' ? `{${insertText}` : insertText
+      insertText = suffix !== '}' ? `${insertText}}` : insertText
+      completionItem.insertText = insertText
+      return completionItem
+    })
```
Mogztter, line 20:
```suggestion
    return Object.entries(attributes).map(([key, value]) => {
      const completionItem = new vscode.CompletionItem({
        label: key,
        description: value,
      }, vscode.CompletionItemKind.Text)
      let insertText = value
      insertText = prefix !== '{' ? `{${insertText}` : insertText
      insertText = suffix !== '}' ? `${insertText}}` : insertText
      completionItem.insertText = insertText
      return completionItem
    })
```

package.json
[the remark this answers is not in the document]
marieflorescontact: I removed the `__metadata` block but every time a run the tests it come back...

Mogztter (approved):
Looks good to me! 👍🏻

Mogztter (approved):
Looks good to me! 👍🏻

PR #583, opened 2022-06-26
Fix call signature for creating preview, closes #582

### LINKED ISSUE #582: Preview on Open to Side appears broken
On the master branch, on my machine, preview using Open to Side appears to be broken.
It only ever opens a preview within the same window I am already using.

Looking at:

https://github.com/asciidoctor/asciidoctor-vscode/blob/e93abba2131c3ab9fbf094fc375d3b4a6e5ab027/src/features/previewManager.ts#L108-L121

Does not seem to match the function signature in:

https://github.com/asciidoctor/asciidoctor-vscode/blob/e93abba2131c3ab9fbf094fc375d3b4a6e5ab027/src/features/preview.ts#L74-L84

`resourceColumn` and `previewColumn` are swapped. I notice that `resourceColumn` is never used according to my linter.

If I swap these it functions correctly.
PR shortly.

danyill:
Closes #582

### DIFF
--- a/src/features/preview.ts
+++ b/src/features/preview.ts
@@ -73,8 +73,8 @@ export class AsciidocPreview extends Disposable implements WebviewResourceProvid
 
   public static create (
     resource: vscode.Uri,
-    previewColumn: vscode.ViewColumn,
     resourceColumn: vscode.ViewColumn,
+    previewColumn: vscode.ViewColumn,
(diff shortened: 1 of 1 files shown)


Mogztter (approved):
👍🏻

marieflorescontact (approved):

PR #587, opened 2022-06-26
resolves #577 normalize setting names and group by category

Mogztter:
- use third-person singular verb form
- introduce 5 categories : preview, pdf, extensions, general, debug
- use order to move most commonly used settings on top (of each category)
- deprecate/rename settings to comply with the new naming convention
- remove unused settings
- update English localization
- use `markdownDeprecationMessage` and `markdownDescription` to format deprecation message and description (using Markdown)
- add `scope: resource` (taken form the Markdown extension)

resolves #577

### TODO

- [x]  Update the Japanese localization
- [x] Add a CHANGELOG entry
- [x] Update code to use setting new names
- [x] Update the logic around the `wkhtmltopdf` command (we need to support command line arguments)
- [x] Update the logic to select a PDF engine
- [x] Update README
- [x] install `asciidoctor-pdf` in the global storage: "A global storage path pointing to a local directory where your extension has read/write access. This is a good option if you need to store large files that are accessible from all workspaces." https://code.visualstudio.com/api/references/vscode-api#ExtensionContext.globalStoragePath
- [x] try to use `asciidoctor-pdf` first and then `bundle exec asciidoctor-pdf` if the `asciidoctor-pdf` command has not been defined (i.e., default value).
    - if the user is using a custom value we should not offer to install `asciidoctor-pdf`

PR #591, opened 2022-07-01
Add Japanese localization for #587

YoshihideShirai:
Hi @Mogztter 

I created Japanese localization for #587.
Would you marge it?


Mogztter:
はい喜んで

PR #594, opened 2022-07-05
fixed for localization issue

YoshihideShirai:
Hi, @Mogztter .

I found a issue that relate a localization .
Localizations didn't work except config's titles or descriptions.

vscode-nls-dev has procedure which is following...
1. Search nls.loadMessageBundle() from js files
2. Create xxx.nls.json by js file name which is found in "1.".
3. Create *.nls.[jp/fr].json from a i18n directory and "2.".

In current master branch, only i18n.ts has nls.loadMessageBundle().
I think that it is a bad implementation. 

Can you check it?
And, please translate English message in i18n/fra/src/security.i18n.json.

### DIFF
--- a/i18n/fra/src/security.i18n.json
+++ b/i18n/fra/src/security.i18n.json
@@ -9,12 +9,11 @@
   "security.disable.description": "Autorise le chargement de l'ensemble des contenus ainsi que l'exécution des scripts. Pas recommandé.",
   "security.enableSecurityWarning.title": "Activer les alertes de sécurité de la prévisualisation sur cet espace de travail",
   "security.disableSecurityWarning.title": "Désactiver les alertes de sécurité de la prévisualisation sur cet espace de travail",
-  "security.toggleSecurityWarning.title": "À noter que cela n'affecte pas le niveau de securité configuré.",
   "security.showPreviewSecuritySelector.title": "Sélectionner le niveau de securité adapté aux prévisualisations des documents AsciiDoc sur cet espace de travail.",
   "security.restrictAsciidoctorExtensionsAuthors.title": "Non approuvé",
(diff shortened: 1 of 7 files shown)


i18n/fra/src/security.i18n.json
```diff
@@ -15,5 +15,6 @@
   "security.restrictAsciidoctorExtensionsAuthors.description": "Désactive l'exécution du code des extensions Asciidoctor.js.",
   "security.trustAsciidoctorExtensionsAuthors.title": "Approuvé",
   "security.trustAsciidoctorExtensionsAuthors.description": "Autorise l'exécution du code des extensions Asciidoctor.js.",
-  "security.asciidoctorExtensionsTrustModeSelector.title": "Séléctionner le niveau de confiance adapté aux extensions Asciidoctor.js présentes sur cet espace de travail."
+  "security.asciidoctorExtensionsTrustModeSelector.title": "Séléctionner le niveau de confiance adapté aux extensions Asciidoctor.js présentes sur cet espace de travail.",
+  "security.toggleSecurityWarning.description": "Please note that it does not affect the content security level."
```
Mogztter, line 19:
```suggestion
  "security.toggleSecurityWarning.description": "À noter que cela n'affecte pas le niveau de securité configuré."
```

Mogztter: I guess `security.toggleSecurityWarning.title` should be removed?

Mogztter:
I think you are right, my goal was to load and configure `nls.config({ messageFormat: nls.MessageFormat.file })()` only once but indeed `nls.rewriteLocalizeCalls` rewrites calls per JavaScript file.

I wonder how it's gonna work if we produce a single JavaScript file: https://github.com/asciidoctor/asciidoctor-vscode/pull/506.

Anyway, I will fetch your branch locally and update the French/English localization.

PR #614, opened 2022-07-28
Resolve Antora resources IDs on images

marieflorescontact:
this is  WIP
I am not sure where to create the contentCatalog (I guess when the Antora support is activated?)


src/asciidocParser.ts
```diff
@@ -125,13 +126,18 @@ export class AsciidocParser {
 
     const registry = processor.Extensions.create()
 
+    const contentCatalog = await getContentCatalog(doc.uri, context)
```
Mogztter, line 129: `contentCatalog` can be undefined

src/asciidoctorWebViewConverter.ts
```diff
@@ -183,6 +187,17 @@ export class AsciidoctorWebViewConverter {
...
     }
+    // if Antora support activated
+    if (nodeName === 'image') {
+      const nodeAttributes = node.getAttributes()
+      const target = nodeAttributes.target
+      const resolvedPath = resolveAntoraImageIds(target, this.contentCatalog, this.src)
+      if (resolvedPath !== undefined) {
+        const alt = node.hasAttribute('alt') ? ` alt="${nodeAttributes.alt.split(':').pop()}"` : ''
+        // TODO check for Antora stylesheet
+        return `<img src= "${resolvedPath}" ${alt}>`
```
Mogztter, line 198: do not implement this part for now, fallback to the `baseConverter`, we will implement this logic later

src/features/antora/antoraSupport.ts
```diff
@@ -83,18 +86,124 @@ export async function getAntoraConfig (textDocument: vscode.TextDocument): Promi
...
+  }
+}
+
+export async function getSrc (textDocumentUri: Uri, contentCatalog: ContentCatalog) {
+  const antoraConfig = await getAntoraConfig(textDocumentUri)
+  const doc = await parseAntoraConfig(textDocumentUri)
+  const contentSourceRootPath = path.dirname(antoraConfig.fsPath)
+  if (doc !== {}) {
+    try {
+      const file = contentCatalog.getByPath({
```
Mogztter, line 159: `contentCatalog` can be undefined

src/features/antora/antoraSupport.ts
```diff
@@ -83,18 +86,124 @@ export async function getAntoraConfig (textDocument: vscode.TextDocument): Promi
...
+          branches: 'HEAD',
+          startPath,
+        }],
+      },
+      runtime: {},
+      site: {},
+    }
+    try {
+      const contentAggregate = await aggregateContent(playbook)
+      return classifyContent(playbook, contentAggregate)
```
Mogztter, line 139: (performance): we should aggregate and classify content only once per playbook. Use event listeners (file added, renamed, removed) to update the contentCatalog.

src/features/antora/antoraSupport.ts
```diff
@@ -83,18 +86,141 @@ export async function getAntoraConfig (textDocument: vscode.TextDocument): Promi
...
+    runtime: {},
+    site: {},
+  }
+}
+
+export async function getSrc (textDocumentUri: Uri, contentCatalog: ContentCatalog | undefined):Promise<{ [key: string]: any } | {}> {
+  const antoraConfig = await getAntoraConfig(textDocumentUri)
+  const doc = await parseAntoraConfig(textDocumentUri)
+  if (antoraConfig !== undefined) {
+    const contentSourceRootPath = path.dirname(antoraConfig.fsPath)
```
Mogztter, line 173: we don't need to compute `contentSourceRootPath` if `doc` is equals to `{}` or if the `contentCatalog` is undefined.
We also don't need to resolve/parse the Antora configuration since we don't have a content catalog.

src/test/asciidoctorWebViewConverter.test.ts
```diff
@@ -0,0 +1,92 @@
...
+</div>`,
+    },
+    // links
+    {
+      title: 'Should resolve macro link',
+      filePath: 'asciidoctorWebViewConverterTest.adoc',
+      input: 'link:full.adoc[]',
+      antoraResourceUri: undefined,
+      expected: `<div class="paragraph">
+<p><a href="full.adoc" class="undefined" data-href="full.adoc">full.adoc</a></p>
```
Mogztter, line 84: It's an existing bug (not related to your changes) but it should be `class="bare"`. I think we should call `node.getRole()` (my guess is that `node.role` is `undefined`)

https://github.com/asciidoctor/asciidoctor-vscode/blob/e454a31d1193d714414a9c399e979a3a217ff188/src/asciidoctorWebViewConverter.ts#L187

Anyway, I will create a dedicated issue.

src/features/previewManager.ts
```diff
@@ -45,10 +45,10 @@ export class AsciidocPreviewManager implements vscode.WebviewPanelSerializer {
     }
   }
 
-  public preview (
+  public async preview (
     resource: vscode.Uri,
     previewSettings: PreviewSettings
-  ): void {
+  ): Promise<void> {
```
Mogztter, line 51: I don't think this method needs to be `async` since we are not using `await` keyword in the method body.

Mogztter:
> Error: src/test/asciidoctorconfig.test.ts(55,108): error TS2345: Argument of type 'WebviewPanel' is not assignable to parameter of type 'WebviewResourceProvider'.
  Type 'WebviewPanel' is missing the following properties from type 'WebviewResourceProvider': asWebviewUri, asMediaWebViewSrc, cspSource

`src/test/asciidoctorconfig.test.ts` is using the `AsciidocParser` so we need to update the call signature.

marieflorescontact:
I use [Sinon.JS](https://sinonjs.org/releases/v14/) as mock library to mock AntoraDocumentContext :

After many tries I didn't manage to test `AsciidoctorWebViewConverter` for an Antora ID. 
`ContentCatalog` was always undefined and `aggregateContent()` sent me an error because the `workspacefolder` used while testing was not a git repository 😕

PR #618, opened 2022-08-14
Search path for included files

eiswind:
I adjusted the search path for included files, in case that the current file is not at the root of the workspace.

I really would like to see that the completions for document attributes are either hidden or shown towards the end of the list, at least they (most of them) do not make much sense in the context of an include. I tried to figure that out, but I could not find the place where the AsciidocProvider gets registered. If I could get some directions here, I'll try to figure that out.

image

### DIFF
--- a/src/providers/asciidoc.provider.ts
+++ b/src/providers/asciidoc.provider.ts
@@ -50,7 +50,6 @@ async function provide (
     }
   }
 
-  const workspace = vscode.workspace.getWorkspaceFolder(context.document.uri)
   const documentPath = context.document.uri.fsPath
   const rootPath = documentPath.substr(0, documentPath.lastIndexOf('/'))
   const searchPath = getPathOfFolderToLookupFiles(
--- a/src/providers/asciidoc.provider.ts
+++ b/src/providers/asciidoc.provider.ts
@@ -87,6 +87,7 @@ async function provide (
 
   return [
     levelUpCompletionItem,
+    ...variablePathSubstitutions,
     ...items.map((child) => {
       const result = createPathCompletionItem(child)
       result.insertText = result.kind === vscode.CompletionItemKind.File ? child.file + '[]' : child.file
@@ -99,7 +100,6 @@ async function provide (
       }
       return result
     }),
-    ...variablePathSubstitutions,
   ]
 }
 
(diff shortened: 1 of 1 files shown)


Mogztter:
Hey, thanks for your contribution!
Could you please open an issue to describe what is the current behavior and what is the desired/expected behavior?

> I really would like to see that the completions for document attributes are either hidden or shown towards the end of the list, at least they (most of them) do not make much sense in the context of an include. I tried to figure that out, but I could not find the place where the AsciidocProvider gets registered. If I could get some directions here, I'll try to figure that out.

I believe you can use/define the `sortText` property: https://stackoverflow.com/questions/56826661/how-to-force-order-in-vscode-extension-completionitems

I think we should do a couple of things:

- group built-in document attributes: https://docs.asciidoctor.org/asciidoc/latest/attributes/document-attributes-ref/ then we can decide which group(s) we want to enable on which context
   - for instance: compliance, localization and numbering attributes don't make much in the context of an include, image, video, link macro.
- when requesting auto-completion on an include, image, video we should prioritize paths to local ressources then user-defined attributes (where the value can be used in a path), then built-in attributes usually used in path then other built-in attributes.
- when requesting auto-completion on the `image` macro, we should prioritize files with image file extensions: svg, png, jpg, jpeg...
- when requesting auto-completion on the `video` macro, we should prioritize files with video file extensions: mp4, mov...
- when requesting auto-completion on the `include` macro, we should only show paths to text files (i.e., remove binary files)

What do you think? We should probably create an issue for each improvements to keep track of things.

eiswind:
Thanks for getting back so quickly.

I created two new issues to get started.

https://github.com/asciidoctor/asciidoctor-vscode/issues/620
https://github.com/asciidoctor/asciidoctor-vscode/issues/619

eiswind:
I'd love to see this little PR merged as well, as it allows to support non-root-folder asciidoc files.

Mogztter:
> I'd love to see this little PR merged as well, as it allows to support non-root-folder asciidoc files.

Sure, I will test your fix locally in the next few days.

Mogztter:
I can confirm that your fix/enhancement is working great!
Could you please fix the linter issue before I merge? You can run `npm run lint` locally. Thanks!

eiswind:
sure!

Issue #621, opened 2022-08-14
show files in front of attributes on include suggestions


eiswind:
I would like to see files show up in front of other attributes on include/image/video autosuggestions.

I would like to set up a PR using the sortOrder attribute.

As @Mogztter  suggested

when requesting auto-completion on an include, image, video we should prioritize paths to local ressources then user-defined attributes (where the value can be used in a path), then built-in attributes usually used in path then other built-in attributes.

when requesting auto-completion on the image macro, we should prioritize files with image file extensions: svg, png, jpg, jpeg...
when requesting auto-completion on the video macro, we should prioritize files with video file extensions: mp4, mov...
when requesting auto-completion on the include macro, we should only show paths to text files (i.e., remove binary files)

when using sortOder this would imply that we set an alphabetical prefix in the various providers that are involved, at least this is what could grasp so far.

Mogztter:
> when using sortOder this would imply that we set an alphabetical prefix in the various providers that are involved, at least this is what could grasp so far.

That's correct. I think using a numeric values is probably better.

eiswind:
Currently an `a_` prefix is in use, but we could change that. I gave this a try to show the files upfront yesterday and it works like a charm. 

I suggest we use

- 10_ for files
- 20_ for doc attributes
- 30_ onwards for built in attributes

are there any other sources that need to be considered?

ggrossetie:
I've made quite a few improvements to autocompletion (next major version). Please open a new issue if there are still points that need improving.

PR #626, opened 2022-08-18
provide sorttext on include suggestions

eiswind:
https://github.com/asciidoctor/asciidoctor-vscode/issues/621

This PR makes files show up in front of attributes on include suggestions. I tried to keep it as simple as possible.

Do you see any cross-side-effects on other scenarios when the `AttributeReferenceProvider' has a sortText?


Mogztter:
> Do you see any cross-side-effects on other scenarios when the `AttributeReferenceProvider' has a sortText?

No, I think it's fine, thanks 👌🏻

Issue #624, opened 2022-08-15
Use the drop into editor API (VS Code 1.70)


Mogztter:
https://code.visualstudio.com/updates/v1_70#_drop-into-editor-api

We should implement this API to create a macro depending on the file type.

marieflorescontact:
👋
I can submit a pull request for this issue!

Mogztter:
Fixed in 02ef77b

PR #627, opened 2022-08-25
add drop images into editor feature

marieflorescontact:
#624

### DIFF
--- a/package.json
+++ b/package.json
@@ -267,6 +267,19 @@
 		"configuration": [
 			{
 				"order": 20,
+				"id": "editor",
+				"title": "%asciidoc.editor.title%",
+				"properties": {
+					"asciidoc.editor.enableDrop": {
+						"type": "boolean",
+						"default": true,
+						"markdownDescription": "%asciidoc.editor.enableDrop.desc%",
+						"scope": "resource"
+					}
+				}
+			},
+			{
+				"order": 21,
 				"id": "preview",
 				"title": "%asciidoc.preview.title%",
 				"properties": {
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -428,6 +428,12 @@
 						"markdownDescription": "%asciidoc.use_asciidoctorpdf.desc%",
 						"markdownDeprecationMessage": "%asciidoc.use_asciidoctorpdf.deprecationMessage%",
 						"scope": "resource"
+					},
+					"asciidoc.editor.drop.enabled": {
```
Mogztter, line 432: I think we should move it into a dedicated group "Editor" since it's not related to the "Preview".
Also, we should name it `asciidoc.editor.enableDrop` to be consistent with other options.

package.nls.fr.json
```diff
@@ -64,5 +64,6 @@
 	"asciidoc.useWorkspaceRoot.deprecationMessage": "Ce paramètre a été remplacé par `#asciidoc.useWorkspaceRootAsBaseDirectory#` et n'a plus aucun effet.",
 	"asciidoc.use_kroki.desc": "**Obsolète:** Active l'extension Kroki afin de générer des diagrammes.",
 	"asciidoc.use_kroki.deprecationMessage": "Ce paramètre a été remplacé par `#asciidoc.extensions.enableKroki#` et n'a plus aucun effet.",
-	"asciidoc.antora.enableAntoraSupport.desc": "Active le support [Antora](https://antora.org/)."
-}
\ No newline at end of file
+	"asciidoc.antora.enableAntoraSupport.desc": "Active le support [Antora](https://antora.org/).",
+	"configuration.asciidoc.editor.drop.enabled": "Active/desactive le glisser-déposer dans l'editeur Asciidoc. Nécessite l'activation `#editor.dropIntoEditor.enabled#`."
```
Mogztter, line 68: Using a `configuration` prefix is a good idea (to avoid conflicts with other translations) but we should do it globally otherwise it's inconsistent. Since that's beyond the scope of this pull request I would rather use  `asciidoc.editor.enableDrop.desc` and prepare another PR if we want to change the naming convention.

src/features/dropIntoEditor.ts
```diff
@@ -0,0 +1,81 @@
...
+
+const imageFileExtensions = new Set<string>([
+  '.bmp',
+  '.gif',
+  '.ico',
+  '.jpe',
+  '.jpeg',
+  '.jpg',
+  '.png',
+  '.psd',
```
Mogztter, line 18: I don't think an HTML page is capable of displaying a preview of a PSD file (but I might be wrong). If that's the case then I would remove this extension.

marieflorescontact: I will try

marieflorescontact: you were right. no PSD...

Issue #380, opened 2020-11-24
.asciidoctorconfig


rdmueller:
When I work with included subdocuments, I often face the problem that the preview can't find my images. 
This is because I set the `:imagesdir:` in the main `.adoc`-File and the included document can't give the preview a hint about this.

A nice solution can be found here: https://intellij-asciidoc-plugin.ahus1.de/docs/users-guide/features/advanced/asciidoctorconfig-file.html

A file called `.asciidoctorconfig` which will be prepended for the preview. This gives you the chance to tell the preview about the correct `imagesdir` and all images will be displayed.

bovinemagnet:
I like this idea, however, can I ask how this is different from including a `_attributes.adoc` file, as this seems to be the common way to do this (I know it is troublesome).

I like it, because it might solve the issue I have with my workflow, which is, I edit with vs-code using asciidoctor-vscode , and publish with Antora, and asciidoc-pdf. 

The two things that stand out to me though are if it is included like it is shown in the example, as a straight `include::` then it feels no different from an `include::_attributes.adoc[]` type approach  (the example is  `include::.asciidoctorconfig[]`)

The second bit, I like, is that there can be more than one existing, I guess if It worked more like a `.gitignore`, where it could  overrides at that point?

So if I think about my workflow, where would the you have an antora project structure such as the following.  There is the opportunity to have the file live within the pages directory (feels like it would have the same issues as I currently face with the `_attributes.adoc` approach), or probably more usefully in each of the specific module directories (ROOT, api, etc).

`
modules\
 ROOT\
   .asciidoctorconfig
   pages\
       .asciidoctorconfig
   partials
   images
api\
   .asciidoctorconfig
   pages\
      .asciidoctorconfig
   partials
   images  
`

thammegowda:
I used `.asciidoctorconfig` to insert some CSS style for HTML preview,
```adoc
ifdef::backend-html5[]
++++
<style>
body {
font-family: "Linux Libertine O", "Helvetica"
}

h1, h2, h3, h4, h5 {
font-family: "Linux Libertine O", "Helvetica"
}
</style>
++++
endif::backend-html5[]
```

this `<style>` tag seems to be inserted before the Level 0 header/title, and it inturn causes
the issue discussed here: https://github.com/asciidoctor/asciidoctor.js/issues/358#issuecomment-317235779 


[...]

PR #636, opened 2022-09-09
Support .asciidoctorconfig at root of the workspace

apupier:
part of #380

demo asciidortorconfigAtRootWorkspaceFolder

TODO for this PR:
- provide test

TODO for another iteration:
- support other places
- support `.asciidoctorconfig.adoc`
- provide a cache mechanism to avoid reading config file every time

Might not have time to work on it before 30th September but wanted to share current state

### DIFF
--- a/src/asciidocParser.ts
+++ b/src/asciidocParser.ts
@@ -1,5 +1,4 @@
 import * as vscode from 'vscode'
-import * as fs from 'fs'
 import * as path from 'path'
 import { AsciidoctorWebViewConverter } from './asciidoctorWebViewConverter'
 import { Asciidoctor } from '@asciidoctor/core'
@@ -8,6 +7,7 @@ import { AsciidocPreviewConfigurationManager } from './features/previewConfig'
 import { SkinnyTextDocument } from './util/document'
 import { IncludeItems } from './asciidoctorFindIncludeProcessor'
 import { AsciidocContributionProvider } from './asciidocExtensions'
+import { getAsciidoctorConfigContent } from './features/asciidoctorConfig'
 
 const asciidoctorFindIncludeProcessor = require('./asciidoctorFindIncludeProcessor')
 
@@ -26,12 +26,25 @@ const previewConfigurationManager = new AsciidocPreviewConfigurationManager()
 export class AsciidocParser {
   private stylesdir: string
   private apsArbiter: AsciidoctorExtensionsSecurityPolicyArbiter
+  public prependExtension: Asciidoctor.Extensions.PreprocessorKlass
 
   constructor (
     readonly contributionProvider: AsciidocContributionProvider,
     readonly aspArbiter: AsciidoctorExtensionsSecurityPolicyArbiter = null,
     private errorCollection: vscode.DiagnosticCollection = null
   ) {
+    this.prependExtension = processor.Extensions.createPreprocessor('PreprendConfigPreprocessorExtension', {
+      postConstruct: function () {
+        this.asciidoctorConfigContent = ''
+      },
+      process: function (doc, reader) {
+        if (this.asciidoctorConfigContent.length > 0) {
+          // otherwise an empty line at the beginning breaks level 0 detection
+          reader.pushInclude(this.asciidoctorConfigContent, undefined, undefined, 1, {})
+        }
+      },
+    }).$new()
+
     // Asciidoctor.js in the browser environment works with URIs however for desktop clients
     // the stylesdir attribute is expected to look like a file system path (especially on Windows)
     if (process.env.BROWSER_ENV) {
@@ -173,16 +186,9 @@ export class AsciidocParser {
     }
 
     try {
-      const workspaceFolder = vscode.workspace.getWorkspaceFolder(doc.uri)
-      if (workspaceFolder !== undefined) {
-        const asciiDoctorConfig = vscode.Uri.joinPath(workspaceFolder.uri, '.asciidoctorconfig')
-        if (fs.existsSync(asciiDoctorConfig.fsPath)) {
-          const asciiDoctorConfigContent = await vscode.workspace.fs.readFile(asciiDoctorConfig)
-          text =
-            ':asciidoctorconfigdir: ' + asciiDoctorConfig.fsPath + '\n\n' +
-            asciiDoctorConfigContent + '\n\n' +
-            text
-        }
+      const asciidoctorConfigContent = await getAsciidoctorConfigContent(doc.uri)
+      if (asciidoctorConfigContent !== undefined) {
+        (this.prependExtension as any).asciidoctorConfigContent = asciidoctorConfigContent
       }
       const document = processor.load(text, options)
       const blocksWithLineNumber = document.findBy(function (b) {
@@ -304,6 +310,7 @@ export class AsciidocParser {
       const kroki = require('asciidoctor-kroki')
       kroki.register(registry)
     }
+    registry.preprocessor(this.prependExtension)
     await this.registerExtensionsInWorkspace(registry)
   }
 
--- a/src/asciidocParser.ts
+++ b/src/asciidocParser.ts
@@ -1,4 +1,5 @@
 import * as vscode from 'vscode'
+import * as fs from 'fs'
 import * as path from 'path'
 import { AsciidoctorWebViewConverter } from './asciidoctorWebViewConverter'
 import { Asciidoctor } from '@asciidoctor/core'
@@ -172,6 +173,17 @@ export class AsciidocParser {
     }
 
     try {
+      const workspaceFolder = vscode.workspace.getWorkspaceFolder(doc.uri)
+      if (workspaceFolder !== undefined) {
+        const asciiDoctorConfig = vscode.Uri.joinPath(workspaceFolder.uri, '.asciidoctorconfig')
+        if (fs.existsSync(asciiDoctorConfig.fsPath)) {
+          const asciiDoctorConfigContent = await vscode.workspace.fs.readFile(asciiDoctorConfig)
+          text =
+            ':asciidoctorconfigdir: ' + asciiDoctorConfig.fsPath + '\n\n' +
+            asciiDoctorConfigContent + '\n\n' +
+            text
+        }
+      }
       const document = processor.load(text, options)
       const blocksWithLineNumber = document.findBy(function (b) {
         return typeof b.getLineNumber() !== 'undefined'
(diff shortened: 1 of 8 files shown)


src/asciidocParser.ts
```diff
@@ -172,6 +173,14 @@ export class AsciidocParser {
...
 
     try {
+      const workspaceFolder = vscode.workspace.getWorkspaceFolder(doc.uri)
+      if (workspaceFolder !== undefined) {
+        const asciiDoctorConfig = vscode.Uri.joinPath(workspaceFolder.uri, '.asciidoctorconfig')
+        if (fs.existsSync(asciiDoctorConfig.fsPath)) {
+          const asciiDoctorConfigContent = await vscode.workspace.fs.readFile(asciiDoctorConfig)
+          text = asciiDoctorConfigContent + '\n' + text
+        }
+      }
```
Mogztter, line 183: I think we should store the `asciidoctorConfig` content in the `workspaceState` to avoid reading a file every time. We should also create a `FileSystemWatcher` to update the content when the file is renamed, deleted or updated: https://code.visualstudio.com/api/references/vscode-api#workspace.createFileSystemWatcher

src/asciidocParser.ts
```diff
@@ -172,6 +173,14 @@ export class AsciidocParser {
     }
 
     try {
+      const workspaceFolder = vscode.workspace.getWorkspaceFolder(doc.uri)
+      if (workspaceFolder !== undefined) {
+        const asciiDoctorConfig = vscode.Uri.joinPath(workspaceFolder.uri, '.asciidoctorconfig')
+        if (fs.existsSync(asciiDoctorConfig.fsPath)) {
+          const asciiDoctorConfigContent = await vscode.workspace.fs.readFile(asciiDoctorConfig)
+          text = asciiDoctorConfigContent + '\n' + text
```
Mogztter, line 181: @ahus1 could you please confirm that this how the `.asciidoctorconfig` content should be used? (i.e., the content is prepended in the AsciiDoc document)

ahus1: Hi @Mogztter - I summarized the handling of the `.asciidoctorconfig` file on this page: https://intellij-asciidoc-plugin.ahus1.de/docs/users-guide/features/advanced/asciidoctorconfig-file.html

What I see in this PR (if I read it correctly), is that this picks a single file in the workspace folder, and prepends it. 

This is a start, still it lacks some of the features used in the IntelliJ plugin. Maybe the additional functionality can be split into futher PRs.

* The file can also be named `.asciidoctorconfig.adoc`, as this enables preprocessor macros if the file would be used as a standard AsciiDoctor include.
* The file can be in any folder, and then it affects all files in its subfolders. So you want to search in all folder leading to the current folder, and they are cascading.
* I found that it would be better to add two newlines here, to avoid problems with stick-to-together-content (won't happen with just attributes, could happen with other content)
* At the beginning of each of the files, I append `:asciidoctorconfigdir: <absolute path filename>`, this comes in handy to setup for example an `imagesdir`. 
* When just prepending the content, I found that the line numbers rendered in warnings will be off by the number of lines prepended. To work around that, I added a preprocessor that calls `reader.push_include`, that won't affect the line numbers.

Code references:

* https://github.com/asciidoctor/asciidoctor-intellij-plugin/blob/64707c7632e1c5e6efc34100dc3cb2dd0bab4ea6/src/main/java/org/asciidoc/intellij/AsciiDocWrapper.java#L710-L745
* https://github.com/asciidoctor/asciidoctor-intellij-plugin/blob/main/src/main/java/org/asciidoc/intellij/asciidoc/PrependConfig.java

Mogztter: > The file can also be named .asciidoctorconfig.adoc, as this enables preprocessor macros if the file would be used as a standard AsciiDoctor include.

I don't understand why `.asciidoctorconfig` or `.asciidoctorconfig.adoc` would be treated differently since we prepend the content of the file. In other words, Asciidoctor is not aware of the file name, no?

> The file can be in any folder, and then it affects all files in its subfolders. So you want to search in all folder leading to the current folder, and they are cascading.


```
app/
├─ .asciidoctorconfig
├─ public/
│  ├─ .asciidoctorconfig
│  ├─ foo.adoc
├─ src/
│  ├─ .asciidoctorconfig
│  ├─ bar.adoc
├─ baz.adoc
```

IIUC, `foo.adoc` will include the content of `app/.asciidoctorconfig` and the content of `public/.asciidoctorconfig` (in that order). Since the content of `public/.asciidoctorconfig` if added after the content of `app/.asciidoctorconfig` it will have a higher precedence?

apupier: Thanks for all these details!

> This is a start, still it lacks some of the features used in the IntelliJ plugin. Maybe the additional functionality can be split into futher PRs.

This is my plan, especially the time that i have to work on the topic. i would like to provide a almost minimal but functional scope first. Objective: provide tested and documented feature for '.asciidoctorconfig' at the root of the workspace without major drawback.

> I found that it would be better to add two newlines here, to avoid problems with stick-to-together-content (won't happen with just attributes, could happen with other content)

Added

> At the beginning of each of the files, I append :asciidoctorconfigdir: <absolute path filename>, this comes in handy to setup for example an imagesdir.

Added

> When just prepending the content, I found that the line numbers rendered in warnings will be off by the number of lines prepended. To work around that, I added a preprocessor that calls reader.push_include, that won't affect the line numbers.

I found that it is added here https://github.com/asciidoctor/asciidoctor-intellij-plugin/blob/95bb8f7cb1e1d0f935f71d4a7e97ba06abe87400/src/main/java/org/asciidoc/intellij/asciidoc/PrependConfig.java#L23 but i'm a bit lost how to register a preprocessor in this code fo rnow. Will continue investigation as it seems required even for a first iteration

> The file can also be named .asciidoctorconfig.adoc, as this enables preprocessor macros if the file would be used as a standard AsciiDoctor include.
> The file can be in any folder, and then it affects all files in its subfolders. So you want to search in all folder leading to the current folder, and they are cascading.

i think these two can be done in separated PRs.

src/asciidocParser.ts
apupier: Unless we detect performance issue, I would prefer to keep it to another iteration. I'm expecting that these files are usually relatively small

apupier:
Updates:

TODO for this PR:
  - provide test

TODO for another iteration:
  -  support other places
  - support .asciidoctorconfig.adoc
  - provide a cache mechanism to avoid reading config file every time

My plan is to look at it again on 7Th ocotober. If someone wants to pick up where i left (or even restart from scratch) to provide the asciidoctorconfig support, this is perfectly fine.

help is welcome to write the test, see my comment:

https://github.com/asciidoctor/asciidoctor-vscode/pull/636/commits/6b2a292bf3af9ade8851424cbff23707c2c7fd66#diff-c425a813b66e595cf78dc31a25bc50fff745cf496eb51fb752c16d62c1798d2bR17

src/asciidocParser.ts
ahus1: +1 for keeping this PR small an concise!

> When just prepending the content, I found that the line numbers rendered in warnings will be off by the number of lines prepended.

I'm not familiar with the VS Code plugin if it (and where) it shows any warnings and errors generated during processing of the preview. If it doesn't this is not needed for the PR. Even if it does, it might still be OK to have this in a future PR IMHO. So I suggest to not see this as a blocker.

ahus1: > I don't understand why .asciidoctorconfig or .asciidoctorconfig.adoc would be treated differently since we prepend the content of the file. In other words, Asciidoctor is not aware of the file name, no?

This makes no different to the IDE as you described. It only makes a different if someone wants to use the file as a standard include in Asciidoctor: Asciidoctor will only process preprocessor macros on includes which have a well-known AsciiDoc extension like ".adoc". 

> IIUC, foo.adoc will include the content of app/.asciidoctorconfig and the content of public/.asciidoctorconfig (in that order). Since the content of public/.asciidoctorconfig if added after the content of app/.asciidoctorconfig it will have a higher precedence?

Sorry for not responding earlier to this: The file higher up in the hierarchy has the highest precedence, wich the file in the root of the project having the highest one. Looking at the final document with all the prepended information. it starts with the file with the highest precedence / the one from the root, and then continues with the files in the folders up to the file wich is pre-rendered.
This is also the way `.editorconfig` works.

Mogztter: > This makes no difference to the IDE as you described. It only makes a different if someone wants to use the file as a standard include in Asciidoctor: Asciidoctor will only process preprocessor macros on includes which have a well-known AsciiDoc extension like ".adoc".

👍🏻 

> Sorry for not responding earlier to this: The file higher up in the hierarchy has the highest precedence, wich the file in the root of the project having the highest one.
> Looking at the final document with all the prepended information. it starts with the file with the highest precedence / the one from the root, and then continues with the files in the folders up to the file wich is pre-rendered.
This is also the way .editorconfig works.

I think we don't have the same definition of higher precedence 😄
But as far as I understand, the `.asciidoctorconfig` file closer to the file takes precedence (i.e., for instance, it will override an attribute defined at the root).

> EditorConfig files are read top to bottom and the most recent rules found take precedence. Properties from matching EditorConfig sections are applied in the order they were read, so properties in closer files take precedence.

In our case, we include the most distant `.asciidoctorconfig` first so that closer `.asciidoctorconfig` fille take precedence.

ahus1: > I think we don't have the same definition of higher precedence

Thanks @Mogztter - you're right and I'm wrong. I stand corrected.

Mogztter:
That's a good plan @apupier 

I fixed the test and introduce a Preprocessor to push the content of the `.asciidoctorconfig` at the start of the document.
As a side note, we desperately need to simplify the code as we have a lot of boilerplate. I believe it will make things easier.

I think it's ready to go. @ahus1 @apupier If you have time to review the code that would be awesome.

src/features/asciidoctorConfig.ts
```diff
@@ -0,0 +1,25 @@
+import * as vscode from 'vscode'
```
apupier, line 1: Should we add the license header?

apupier (commented):
tested locally and it works fine.

Thank you very much for the addition of the test and the preprocessor!

(I cannot approve as I created the Pull request)

src/features/asciidoctorConfig.ts
Mogztter: To be honest, I want to remove the license header everywhere. In my opinion, it adds unnecessary noise.
The extension is based on https://github.com/microsoft/vscode/tree/main/extensions/markdown-language-features but the license is MIT so https://github.com/asciidoctor/asciidoctor-vscode/blob/master/LICENSE should be enough.

PR #646, opened 2022-10-07
Support .asciidoctorconfig.adoc at root of the workspace folder

apupier:
part of #380

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Improvements
 
-- support .asciidoctorconfig at root of the workspace by @apupier (#380)
+- support `.asciidoctorconfig` and `.asciidoctorconfig.adoc` at root of the workspace by @apupier, @mogztter and @ahus1 (#380)
 
 ### Bug fixes
 
--- a/src/features/asciidoctorConfig.ts
+++ b/src/features/asciidoctorConfig.ts
@@ -17,9 +17,25 @@ export async function getAsciidoctorConfigContent (documentUri: vscode.Uri): Pro
   if (workspaceFolder === undefined) {
     return undefined
   }
-  const asciidoctorConfigUri = vscode.Uri.joinPath(workspaceFolder.uri, '.asciidoctorconfig')
+  const configContentFromAsciidoctorConfig = await getConfigContent(workspaceFolder, '.asciidoctorconfig')
+  const configContentFromAsciidoctorConfigDotAdoc = await getConfigContent(workspaceFolder, '.asciidoctorconfig.adoc')
+
+  const configContents = [
+    configContentFromAsciidoctorConfig,
+    configContentFromAsciidoctorConfigDotAdoc,
+  ].filter((config) => config !== undefined)
+
+  if (configContents.length > 0) {
+    return configContents.join('\n\n')
+  }
+  return undefined
+}
+
+async function getConfigContent (workspaceFolder: vscode.WorkspaceFolder, configFilename: string) {
+  const asciidoctorConfigUri = vscode.Uri.joinPath(workspaceFolder.uri, configFilename)
   if (await exists(asciidoctorConfigUri)) {
     const asciidoctorConfigContent = new TextDecoder().decode(await vscode.workspace.fs.readFile(asciidoctorConfigUri))
     return `${asciidoctorConfigContent.trim()}\n\n`
   }
+  return undefined
 }
--- a/README.md
+++ b/README.md
@@ -172,7 +172,7 @@ To provide a common set of variables when rendering the preview, the extension r
 
 It is inspired by the implementation provided in [IntelliJ AsciiDoc Plugin](https://intellij-asciidoc-plugin.ahus1.de/docs/users-guide/features/advanced/asciidoctorconfig-file.html) and reused in [Eclipse AsciiDoc plugin](https://github.com/de-jcup/eclipse-asciidoctor-editor/wiki/Asciidoctor-configfiles).
 
-The current scope is limited to an `.asciidoctorconfig` configuration file provided at the root of the workspace.
+The current scope is limited to an `.asciidoctorconfig` or `.asciidoctorconfig.adoc` configuration file provided at the root of the workspace.
 
 ## Extension Settings
 
--- a/src/test/asciidoctorconfig.test.ts
+++ b/src/test/asciidoctorconfig.test.ts
@@ -68,24 +68,26 @@ suite('asciidoc.Asciidoctorconfig', () => {
     })
   })
 
-  test('Pick up .asciidoctorconfig and .asciidoctorconfig.adoc from root workspace folder', async () => {
-    const webview = vscode.window.createWebviewPanel(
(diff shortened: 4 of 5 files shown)


src/features/asciidoctorConfig.ts
```diff
@@ -17,9 +17,19 @@ export async function getAsciidoctorConfigContent (documentUri: vscode.Uri): Pro
   if (workspaceFolder === undefined) {
     return undefined
   }
-  const asciidoctorConfigUri = vscode.Uri.joinPath(workspaceFolder.uri, '.asciidoctorconfig')
+  const configContent = await getConfigContent(workspaceFolder, '.asciidoctorconfig')
+  if (configContent !== undefined) {
+    return configContent
+  } else {
+    return getConfigContent(workspaceFolder, '.asciidoctorconfig.adoc')
+  }
```
Mogztter, line 25:
```suggestion
  return getConfigContent(workspaceFolder, '.asciidoctorconfig.adoc')
```

src/features/asciidoctorConfig.ts
```diff
@@ -17,9 +17,19 @@ export async function getAsciidoctorConfigContent (documentUri: vscode.Uri): Pro
   if (workspaceFolder === undefined) {
     return undefined
   }
-  const asciidoctorConfigUri = vscode.Uri.joinPath(workspaceFolder.uri, '.asciidoctorconfig')
+  const configContent = await getConfigContent(workspaceFolder, '.asciidoctorconfig')
```
Mogztter, line 20: I might be wrong but it seems that the Intellij plugin will return the content of both files. So if you have:

**.asciidoctorconfig**
```
:name: Aurélien
:version: 1.2.3
```
**.asciidoctorconfig.adoc**
```
:name: Alexander
```

It will include the following contents:

```
:name: Aurélien
:version: 1.2.3


:name: Alexander


```

https://github.com/asciidoctor/asciidoctor-intellij-plugin/blob/64707c7632e1c5e6efc34100dc3cb2dd0bab4ea6/src/main/java/org/asciidoc/intellij/AsciiDocWrapper.java#L710-L745

So `.asciidoctorconfig.adoc` has higher precedence and both files are taken into account.

Mogztter: @ahus1 is that correct?

ahus1: @Mogztter - yes, you're right, this is the order it is being processed in. 

I doubt that a user should have both files in a folder, still it makes sense to have a consistent behavior across tools.

Thanks for paying extra attention here!

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 
 ### Improvements
 
-- support .asciidoctorconfig at root of the workspace by @apupier (#380)
+- support `.asciidoctorconfig` and `.asciidoctorconfig.adoc` at root of the workspace by @apupier , @mogztter and @ahus1 (#380)
```
ahus1, line 7: In English text, there shouldn't be a blank before the comma.

```suggestion
- support `.asciidoctorconfig` and `.asciidoctorconfig.adoc` at root of the workspace by @apupier, @mogztter and @ahus1 (#380)
```

src/features/asciidoctorConfig.ts
apupier: Thanks for pointing it out. I updated the Pull Request to have the same behavior

apupier:
@Mogztter and @ahus1 Please do a second review. i incorporated the feedback. And sorry for pining instead of only requesting a review but GitHub is drunk, as soon as I ask for a review on one of you, it dismiss the ask for review on the other...

src/features/asciidoctorConfig.ts
```diff
@@ -17,9 +17,28 @@ export async function getAsciidoctorConfigContent (documentUri: vscode.Uri): Pro
   if (workspaceFolder === undefined) {
     return undefined
   }
-  const asciidoctorConfigUri = vscode.Uri.joinPath(workspaceFolder.uri, '.asciidoctorconfig')
+  const configContentFromAsciidoctorConfig = await getConfigContent(workspaceFolder, '.asciidoctorconfig')
+  const configContentFromAsciidoctorConfigDotAdoc = await getConfigContent(workspaceFolder, '.asciidoctorconfig.adoc')
+  if (configContentFromAsciidoctorConfig !== undefined) {
+    if (configContentFromAsciidoctorConfigDotAdoc !== undefined) {
+      return configContentFromAsciidoctorConfig + configContentFromAsciidoctorConfigDotAdoc
```
Mogztter, line 24: The Intellij IDEA plugin joins configuration files using two newlines:

https://github.com/asciidoctor/asciidoctor-intellij-plugin/blob/64707c7632e1c5e6efc34100dc3cb2dd0bab4ea6/src/main/java/org/asciidoc/intellij/AsciiDocWrapper.java#L727

src/features/asciidoctorConfig.ts
```diff
@@ -17,9 +17,28 @@ export async function getAsciidoctorConfigContent (documentUri: vscode.Uri): Pro
   if (workspaceFolder === undefined) {
     return undefined
   }
-  const asciidoctorConfigUri = vscode.Uri.joinPath(workspaceFolder.uri, '.asciidoctorconfig')
+  const configContentFromAsciidoctorConfig = await getConfigContent(workspaceFolder, '.asciidoctorconfig')
+  const configContentFromAsciidoctorConfigDotAdoc = await getConfigContent(workspaceFolder, '.asciidoctorconfig.adoc')
```
Mogztter, line 21: To simplify the code we can do the following:

1. Create an array of configuration files
2.  Filter `undefined` values in the array
3. If the array is empty, return `undefined`
4. Otherwise, join the array using `\n\n`

Pseudo-code (untested)
```ts
const configContentFromAsciidoctorConfig = await getConfigContent(workspaceFolder, '.asciidoctorconfig')
const configContentFromAsciidoctorConfigDotAdoc = await getConfigContent(workspaceFolder, '.asciidoctorconfig.adoc')

const configContents = [
  configContentFromAsciidoctorConfig,
  configContentFromAsciidoctorConfigDotAdoc
].filter(config => config !== undefined)

if (configContents.length > 0) {
  return configContents.join(`\n\n`)
}
return undefined
```

src/features/asciidoctorConfig.ts
apupier: sounds overkill as we already have two newlines at end of the grabbed content but well, better to stay as close as possible to IntelliJ

src/features/asciidoctorConfig.ts
apupier: I find it clearer having the 4 cases in different if/else cases than using a loop but the end result is the same and higher chance that you will have to maintain it than me. I updated with your proposal.

ahus1:
Hi, unfortunately I didn't find the time yet to look at this PR again; will try my best to do this later this week. Sorry for the delay. Feel free to proceed without my input, I don't want to block your efforts here.

Mogztter (approved):
It looks good, thanks!

Issue #648, opened 2022-10-14
Autocompletion for references


jens-auer:
Similar to autocompletion for image and include files (#377), I would like to use autocompletion for internal and external references. 

Details:
- Start typing << and an autosuggest should offer IDs in the current document
- Start typing xref: and an autosuggest should offer asciidoc files in the project. If the user selects a file foo.adoc, it should autocomplete to xref:foo.adoc# and then autosuggest IDs in the target document

Background:
We import requirements from a requirement management tool and reference them in the text. Autocomplete would reduce errors and provide better efficiency.

Mogztter:
Are you talking specifically about Antora xref or AsciiDoc xref?

jens-auer:
AsciiDoc xref (https://docs.asciidoctor.org/asciidoc/latest/macros/inter-document-xref)

kharann:
Agreed! This would be a nice feature :)))

apupier:
The IDs that can be referenced are these ones? https://docs.asciidoctor.org/asciidoc/latest/attributes/id/

@Mogztter Do we already have an utility parsing.adoc to retrieve all IDs of a document?

I think that we can split this task with:
-  [ ] Create utility method to retrieve IDs in a document
   - [ ] for simple case, only id defined in square bracket with shorthand syntax `#`
   - [ ] for simple case, only id defined in square bracket with longhand syntax `id=`
   - [ ]  for simple case, only id defined in double square brackets with legacy syntax
   - [ ] for most advanced case when there are other things in square brackets like roles [quote.movie#roads,Dr. Emmett Brown]
- [ ] Provide completion in the same document after typing <<
- [ ] Propose completion in the same document when starting a word proposing the <<id>> ?
- [ ] Provide completion for other document in the workspace when typing xref: (should we restrict to .adoc document?) But maybe this one will need to be discussed a bit more because I feel like there are different ways to do it https://docs.asciidoctor.org/pdf-converter/latest/interdocument-xrefs/
- [ ] Provide completion for ids in referenced document when typing `xref:my.adoc#`

Mogztter:
> Do we already have an utility parsing .adoc files to retrieve all IDs of a document?

I guess you can [`load`](https://github.com/asciidoctor/asciidoctor-vscode/blob/bf5ee4fffd8889f299fdca2255080c62bf022368/src/asciidocParser.ts#L99-L118) the document and retrieve anchors using the catalog:

[...]

PR #667, opened 2022-11-18
Fix completion after xref: for old double-square bracket notation

apupier:
part of #648

- it was broken during enforcing code style, the regular expression was not correctly converted
https://github.com/asciidoctor/asciidoctor-vscode/commit/cd477c7
- added a test
- Maybe not very nice in the sense that it is not inserting the file path but only the label, but maybe it is enough if there is the right include?

image

### DIFF
--- a/src/providers/xref.provider.ts
+++ b/src/providers/xref.provider.ts
@@ -30,14 +30,17 @@ function shouldProvide (context: Context): boolean {
 }
 
 async function getLabels (): Promise<string[]> {
-  const regex = /\\[\\[(\\w+)\\]\\]/g
-  const labels = await vscode.workspace.findFiles('**/*.adoc').then((files) =>
-    files
+  const regex = /\[\[(\w+)\]\]/g
+  const labels = await vscode.workspace.findFiles('**/*.adoc').then((files) => {
+    const contentOfFilesConcatenated = files
       .map((uri) => readFileSync(uri.path).toString('utf-8'))
       .join('\n')
-      .match(regex)
-      .map((result) => result.replace('[[', '').replace(']]', ''))
-  )
+    const matched = contentOfFilesConcatenated.match(regex)
+    if (matched) {
+      return matched.map((result) => result.replace('[[', '').replace(']]', ''))
+    }
+    return []
+  })
   return labels
 }
 
--- a/src/test/xrefCompletionProvider.test.ts
+++ b/src/test/xrefCompletionProvider.test.ts
@@ -1,0 +1,36 @@
+import 'mocha'
+import * as vscode from 'vscode'
+import assert from 'assert'
+import { xrefProvider } from '../providers/xref.provider'
+import { Position } from 'vscode'
+
+let root
+
+suite('Xref CompletionsProvider', () => {
+  const createdFiles: vscode.Uri[] = []
+  setup(() => {
+    root = vscode.workspace.workspaceFolders[0].uri.fsPath
+  })
+  teardown(async () => {
+    for (const createdFile of createdFiles) {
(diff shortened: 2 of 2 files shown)


src/providers/xref.provider.ts
```diff
@@ -30,14 +30,17 @@ function shouldProvide (context: Context): boolean {
...
+  const regex = /\[\[(\w+)\]\]/g
+  const labels = await vscode.workspace.findFiles('**/*.adoc').then((files) => {
+    const contentOfFilesConcatenated = files
       .map((uri) => readFileSync(uri.path).toString('utf-8'))
       .join('\n')
-      .match(regex)
-      .map((result) => result.replace('[[', '').replace(']]', ''))
-  )
+    const matched = contentOfFilesConcatenated.match(regex)
+    if (matched) {
```
apupier, line 39: checking for null, previously there was an exception caught by the VS Code framework.

src/providers/xref.provider.ts
```diff
@@ -30,14 +30,17 @@ function shouldProvide (context: Context): boolean {
 }
 
 async function getLabels (): Promise<string[]> {
-  const regex = /\\[\\[(\\w+)\\]\\]/g
-  const labels = await vscode.workspace.findFiles('**/*.adoc').then((files) =>
-    files
+  const regex = /\[\[(\w+)\]\]/g
+  const labels = await vscode.workspace.findFiles('**/*.adoc').then((files) => {
+    const contentOfFilesConcatenated = files
       .map((uri) => readFileSync(uri.path).toString('utf-8'))
```
Mogztter, line 36: Not related but I think we should use https://code.visualstudio.com/api/references/vscode-api#FileSystem to `readFile`.

As far as I know, `readFileSync` won't be available/won't work in VS Code Web.

Mogztter:
> Maybe not very nice in the sense that it is not inserting the file path but only the label, but maybe it is enough if there is the right include?

We can use `insertText` on completionItems to provide a text that should be inserted in the document when selection a completion: https://code.visualstudio.com/api/references/vscode-api#CompletionItem

Mogztter:
> added a test

THANKS 🙌🏻

src/providers/xref.provider.ts
apupier: agree.
I only reenabled what was already implemented.
I can change this Pull Request or update in another iteration.

(For information, planned to work on it on Friday 25th October)

Mogztter: > I can change this Pull Request or update in another iteration.

I'm fine with both!

apupier: I would prefer to it in a seprate Pull request. I listed the task in https://github.com/asciidoctor/asciidoctor-vscode/issues/648#issuecomment-1327445736

apupier:
> > Maybe not very nice in the sense that it is not inserting the file path but only the label, but maybe it is enough if there is the right include?
> 
> We can use `insertText` on completionItems to provide a text that should be inserted in the document when selection a completion: https://code.visualstudio.com/api/references/vscode-api#CompletionItem

My question was if it is even valid to insert it this way currently. But I guess we can split it anyway. it is helping users anyway, and can insert in a second iteration. I mentioned it as a seprate task here https://github.com/asciidoctor/asciidoctor-vscode/issues/648#issuecomment-1327445736

apupier:
@Mogztter if fine for you, I think we can merge this Pull Request as-is and iterate to improve the mentioned points.

Mogztter:
@apupier And merged, thanks!

You will need to rebase your other pull requests 😬

PR #668, opened 2022-11-18
ref #648 provide completion short hand and long hand notation with similar scope of legacy

apupier:
Provide completion after xref for ids defined with shorthand and
longhand notation

I.e. with [#myId] and [id=myId]

part of #648

image

based on https://github.com/asciidoctor/asciidoctor-vscode/pull/667

### DIFF
--- a/src/providers/xref.provider.ts
+++ b/src/providers/xref.provider.ts
@@ -30,18 +30,41 @@ function shouldProvide (context: Context): boolean {
 }
 
 async function getLabels (): Promise<string[]> {
+  const files = await vscode.workspace.findFiles('**/*.adoc')
+  const contentOfFilesConcatenated = files
+    .map((uri) => readFileSync(uri.path).toString('utf-8'))
+    .join('\n')
+  const labelsFromLegacyBlock = await getLabelsFromLegacyBlock(contentOfFilesConcatenated)
+  const labelsFromShorthandNotation = await getLabelsFromShorthandNotation(contentOfFilesConcatenated)
+  const labelsFromLonghandNotation = await getLabelsFromLonghandNotation(contentOfFilesConcatenated)
+  return labelsFromLegacyBlock.concat(labelsFromShorthandNotation, labelsFromLonghandNotation)
+}
+
+async function getLabelsFromLonghandNotation (content: string): Promise<string[]> {
+  const regex = /\[id=(\w+)\]/g
+  const matched = content.match(regex)
+  if (matched) {
+    return matched.map((result) => result.replace('[id=', '').replace(']', ''))
+  }
+  return []
+}
+
+async function getLabelsFromShorthandNotation (content: string): Promise<string[]> {
+  const regex = /\[#(\w+)\]/g
+  const matched = content.match(regex)
+  if (matched) {
+    return matched.map((result) => result.replace('[#', '').replace(']', ''))
+  }
+  return []
+}
+
+async function getLabelsFromLegacyBlock (content: string): Promise<string[]> {
   const regex = /\[\[(\w+)\]\]/g
-  const labels = await vscode.workspace.findFiles('**/*.adoc').then((files) => {
-    const contentOfFilesConcatenated = files
-      .map((uri) => readFileSync(uri.path).toString('utf-8'))
-      .join('\n')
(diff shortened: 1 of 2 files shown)


src/providers/xref.provider.ts
```diff
@@ -30,15 +30,41 @@ function shouldProvide (context: Context): boolean {
...
+    .map((uri) => readFileSync(uri.path).toString('utf-8'))
+    .join('\n')
+  const labelsFromLegacyBlock = await getLabelsFromLegacyBlock(contentOfFilesConcatenated)
+  const labelsFromShorthandNotation = await getLabelsFromShorthandNotation(contentOfFilesConcatenated)
+  const labelsFromLonghandNotation = await getLabelsFromLonghandNotation(contentOfFilesConcatenated)
+  return labelsFromLegacyBlock.concat(labelsFromShorthandNotation, labelsFromLonghandNotation)
+}
+
+async function getLabelsFromLonghandNotation (content: string): Promise<string[]> {
+  const regex = /\[id=(\w+)\]/g
```
Mogztter, line 44: I think we need to check that the line starts with `[` using `^`

src/providers/xref.provider.ts
```diff
@@ -30,15 +30,41 @@ function shouldProvide (context: Context): boolean {
...
+  const matched = content.match(regex)
+  if (matched) {
+    return matched.map((result) => result.replace('[id=', '').replace(']', ''))
```
Mogztter, line 47: we should probably ignore everything after the `]` or make sure that the line ends with `]`

Mogztter, line 47: Might be better to use the (named) group? Since we are using `(\w+)` we can use the group reference `$1`

apupier: right, it is necessary to support inlined anchor https://docs.asciidoctor.org/asciidoc/latest/attributes/id/#as-an-inline-anchor

src/providers/xref.provider.ts
apupier: Seems there is a case where it is not at the begiining of the line, for inlined anchor:
https://docs.asciidoctor.org/asciidoc/latest/attributes/id/#as-an-inline-anchor

```
* First item
* [[step2]]Second item
* Third item
```

src/providers/xref.provider.ts
apupier: also for additional anchors to a section https://docs.asciidoctor.org/asciidoc/latest/attributes/id/#add-additional-anchors-to-a-section

src/providers/xref.provider.ts
apupier: also for additional anchors to a section https://docs.asciidoctor.org/asciidoc/latest/attributes/id/#add-additional-anchors-to-a-section

```
[#version-4_9]
=== [[current]][[latest]]Version 4.9
```

src/providers/xref.provider.ts
apupier: in fact, it is already working because ti is using the matched string, so there is nothign after the `]` I added a test case.

image

apupier: seems we need a more recent version of TypeScript https://javascript.info/regexp-groups#searching-for-all-matches-with-groups-matchall

src/providers/xref.provider.ts
Mogztter: You are right... seems like we need a better parsing strategy if we want to be 100% accurate.

src/providers/xref.provider.ts
Mogztter: `matchAll` is available on Node 12. Are you sure it does not work?

apupier: I have this error:
image

Mogztter: You can safely update the target to `es2020`.

I'm planning to make the switch: https://github.com/asciidoctor/asciidoctor-vscode/pull/506

apupier: I would prefer that we keep it in a separate Pull request. I'm not even sure how to change the target correctly neither what are the implications

Mogztter: All good 👍🏻

PR #669, opened 2022-11-25
ref #648 use vscode fs api instead of Node fs

apupier:
it will allow to compliant with VS Code web

part of #648

based on https://github.com/asciidoctor/asciidoctor-vscode/pull/668 to ease rebase and avoid conflicts


Mogztter:
Thanks! 

(et bonnes fêtes 🎄)

PR #670, opened 2022-11-25
648 provide completion after<<from same document

apupier:
part of #648 

Screenshot from 2022-11-25 16-24-09

based on https://github.com/asciidoctor/asciidoctor-vscode/pull/669


ggrossetie (approved):
It looks good, thanks! 🙌🏻

PR #671, opened 2022-11-25
648 Include path to completion item for xref

apupier:
one test provided when at same level

image

part of #648

would be nice to add a test for files at different levels. i think this can be provided in another pull request. I do not know when I will have time to come back to it


ggrossetie:
Thanks @apupier 🙌🏻

limscoder:
Hi, thanks so much for working on this feature, I'm looking forward to it. When can we expect a release version with this feature available?

ggrossetie:
@limscoder I've released a new pre-version 3.0.5, please give it a try!

Issue #678, opened 2022-12-16
Set Up "allow kroki-server-url by default" or add new documentation for "Asciidoc: Change Preview Security Settings" option


haydencbarnes:
After following the same setup as 'willfe' on #628 - setting up a self-hosted Kroki instance utilizing all the documentation available and multiple git issues, I found that the only reason my graphs were not rendering was due to the "Asciidoc: Change Preview Security Settings" option needing to be set correctly to allow the local content to load. See below:

![image](https://user-images.githubusercontent.com/120114860/208027295-c89c827e-ae98-419f-b10c-86626782115b.png)

Note that using a remote server (https://kroki.io/) is allowed by default. So it seems fair that we have the ability to also allow kroki-server-url by default.

Although if documentation is the only answer right now, this may go better on this documentation here: https://github.com/Mogztter/asciidoctor-kroki#using-your-own-kroki. (Although it is an Asciidoc setting, not Kroki.)

Specifically, next to this sentence: "For instance, if you have followed [the instructions](https://docs.kroki.io/kroki/setup/install/#_using_docker) to set up a self-managed server using Docker, you can use the following: `:kroki-server-url: http://localhost:8080`. **+** Along with changing the required "Asciidoc: Change Preview Security Settings" option in VSCode to "Allow insecure local content" to allow the local HTTP content on the server to load in VSCode.

So if we wanted to make that section more structured and direct:

> If you have followed the instructions to set up a self-managed server using Docker, you will need to:
> 
> - Use the following `kroki-server-url: http://localhost:8080/`
> - In VSCode, go to the "Asciidoc: Change Preview Security Settings" option and select "Allow insecure local content" to allow the server's content to load in VSCode.
> 
> This will allow you to use the server and view its content in VSCode.
>

haydencbarnes:
Referencing [https://github.com/asciidoctor/asciidoctor-vscode/issues/628#issue-1353339880](url)

haydencbarnes:
#628 - Reference

haydencbarnes:
@Mogztter Would you like me to submit this edit as a PR on the readme?

Mogztter:
I would rather fix this issue.
I don't want to add it to the README since it's a "quick fix" in:

https://github.com/asciidoctor/asciidoctor-vscode/blob/be5055d2f5f314c865f0b6cd597d2714e0802f4c/src/asciidoctorWebViewConverter.ts#L58

[...]

PR #680, opened 2022-12-22
#678 - Add VSCode rule to allow kroki-server-url by default for local server content

haydencbarnes:
#678 - reference


src/asciidoctorWebViewConverter.ts
```diff
@@ -69,9 +69,15 @@ function getCspForResource (webviewResourceProvider: WebviewResourceProvider, se
       return '<meta http-equiv="Content-Security-Policy" content="">'
 
     case AsciidocPreviewSecurityLevel.Strict:
-    default:
-      return `<meta http-equiv="Content-Security-Policy" content="default-src 'none'; img-src 'self' ${rule} https: data:; media-src 'self' ${rule} https: data:; script-src 'nonce-${nonce}' '${highlightjsInlineScriptHash}'; style-src 'self' ${rule} 'unsafe-inline' https: data:; font-src 'self' ${rule} https: data:;">`
-  }
+      // Initialize the cspRule variable with the default CSP rule
+      let cspRule = `default-src 'none'; img-src 'self' ${rule} https: data:; media-src 'self' ${rule} https: data:; script-src 'nonce-${nonce}' '${highlightjsInlineScriptHash}'; style-src 'self' ${rule} 'unsafe-inline' https: data:; font-src 'self' ${rule} https: data:;`;
+      // If asciidoc.extensions.enableKroki is true, add the kroki-server-url attribute to the CSP rule
+      if (asciidoc.extensions.enableKroki) {
```
Mogztter, line 75: Unless I'm missing something `asciidoc` does not exist in this context?
I think we need to add `enableKroki` and `krokiServerUrl` to the `AsciidocPreviewConfiguration`, then pass the configuration to the `getCspForResource` method

Mogztter: I also think we should add this logic in `AsciidocPreviewSecurityLevel.AllowInsecureLocalContent`

src/asciidoctorWebViewConverter.ts
```diff
@@ -69,9 +69,15 @@ function getCspForResource (webviewResourceProvider: WebviewResourceProvider, se
       return '<meta http-equiv="Content-Security-Policy" content="">'
 
     case AsciidocPreviewSecurityLevel.Strict:
-    default:
-      return `<meta http-equiv="Content-Security-Policy" content="default-src 'none'; img-src 'self' ${rule} https: data:; media-src 'self' ${rule} https: data:; script-src 'nonce-${nonce}' '${highlightjsInlineScriptHash}'; style-src 'self' ${rule} 'unsafe-inline' https: data:; font-src 'self' ${rule} https: data:;">`
-  }
+      // Initialize the cspRule variable with the default CSP rule
+      let cspRule = `default-src 'none'; img-src 'self' ${rule} https: data:; media-src 'self' ${rule} https: data:; script-src 'nonce-${nonce}' '${highlightjsInlineScriptHash}'; style-src 'self' ${rule} 'unsafe-inline' https: data:; font-src 'self' ${rule} https: data:;`;
```
Mogztter, line 73: As a side note, we might want to use https://www.npmjs.com/package/content-security-policy-builder (or similar) to build CSP and avoid duplications/mistakes/typos.

If we don't want to use a library we could at least `join` to make it easier to read:

```js
const rules = [
  "default-src 'none'",
  `img-src 'self' ${rule} https: data:`,
  `media-src 'self' ${rule} https: data:`,
  `script-src 'nonce-${nonce}' '${highlightjsInlineScriptHash}'`,
  `style-src 'self' ${rule} 'unsafe-inline' https: data:`,
  `font-src 'self' ${rule} https: data:`
]
rules.join('; ')
```

Don't make this change in this pull request, that's just an idea to make it easier to maintain in the future.

src/asciidoctorWebViewConverter.ts
haydencbarnes: @Mogztter, why do we need to add this logic to `AsciidocPreviewSecurityLevel.AllowInsecureLocalContent` when it will now be the default?

haydencbarnes: This is already resolved when localhost is allowed:
```
case AsciidocPreviewSecurityLevel.AllowInsecureLocalContent:
    return `<meta http-equiv="Content-Security-Policy" content="default-src 'none'; img-src 'self' ${rule} https: data: http://localhost:* http://127.0.0.1:*; media-src 'self' ${rule} https: data: http://localhost:* http://127.0.0.1:*; script-src 'nonce-${nonce}' '${highlightjsInlineScriptHash}'; style-src 'self' ${rule} 'unsafe-inline' https: data: http://localhost:* http://127.0.0.1:*; font-src 'self' ${rule} https: data: http://localhost:* http://127.0.0.1:*;">`
```

haydencbarnes: https://github.com/Mogztter/asciidoctor-kroki#using-your-own-kroki Notes that "Note that either the http:// or https:// prefix is required (the default Docker image only uses HTTP)." - so I am not sure about adding the logic to .AllowInsecureLocalContent

haydencbarnes: > Unless I'm missing something `asciidoc` does not exist in this context? I think we need to add `enableKroki` and `krokiServerUrl` to the `AsciidocPreviewConfiguration`, then pass the configuration to the `getCspForResource` method

It exists in the Asciidoc parser -   https://github.com/asciidoctor/asciidoctor-vscode/blob/823bb2262a0072a3999ea2f6c15d8794ed92a3df/src/asciidocParser.ts#L318

haydencbarnes: If `enableKroki` exists only at the parser level, should we not just add the default to the parser file, something like this - 
``` private async registerAsciidoctorExtensions (registry, cspRule: string) {
  const enableKroki = vscode.workspace.getConfiguration('asciidoc.extensions', null).get('enableKroki')
  if (enableKroki) {
    const kroki = require('asciidoctor-kroki')
    kroki.register(registry)
    cspRule += `; img-src 'self' ${kroki-server-url} data:;`
  }
  registry.preprocessor(this.prependExtension)
  await this.registerExtensionsInWorkspace(registry, cspRule)
}
```

haydencbarnes: Or this problem is much simpler than it seems. Maybe all I need to do is change the security level once enableKroki is equal to true? - 
```
const enableKroki = vscode.workspace.getConfiguration('asciidoc.extensions', null).get('enableKroki')
if (enableKroki) {
  securityLevel = AsciidocPreviewSecurityLevel.AllowInsecureLocalContent;
  const kroki = require('asciidoctor-kroki')
  kroki.register(registry)
}
registry.preprocessor(this.prependExtension)
await this.registerExtensionsInWorkspace(registry)
}
```

Mogztter: > why do we need to add this logic to AsciidocPreviewSecurityLevel.AllowInsecureLocalContent when it will now be the default?

Because the Kroki server URL should always be allowed no matter which security level is selected. The Kroki server URL can be any URL:

- `https://kroki.io`
- `http://localhost:8000`
- `https://example.org/kroki`
- `http://kroki.internal:1234` 


> If enableKroki should only change the security level when kroki-server-url has been changed, we can declare a variable - const krokiServerUrl = previewAttributes['kroki-server-url'] and then change the if statement to include a check if kroki-server-url has been set?

I think it's easier to always allow the `kroki-server-url`.

> Instead of automatically adding a rule when enableKroki is true, maybe all I need to do is change the security level once enableKroki is equal is true? Is that allowed? (It's logically the same thing happening if the default for the kroki.io extension is sending information and receives diagrams back from https//kroki.io, the user is deciding to enable the graphs being rendered elsewhere than the VSCode space) -

I don't think we should change the security level but when `enableKroki` is true we should allow the `kroki-server-url` not matter which security level is selected.

Does it make sense?

haydencbarnes: Okay, to make it easier to always allow the kroki-server-url, I would refactor the code under the `getCspForResource` function to be structured as cspRules, and then add this if statement to append the cspRule no matter which security level is selected:
```
// Check if the enableKroki configuration is set to true
  if (vscode.workspace.getConfiguration('asciidoc.extensions', null).get('enableKroki')) {
    // Append the kroki-server-url to the CSP rule
    cspRule += `; img-src 'self' ${kroki-server-url};`
  }
```

Mogztter: Sounds good 👍🏻

haydencbarnes: @Mogztter Plz review when you have time new commit on this branch for these edits

src/asciidoctorWebViewConverter.ts
```diff
@@ -58,20 +58,33 @@ const previewStrings = {
...
-    default:
-      return `<meta http-equiv="Content-Security-Policy" content="default-src 'none'; img-src 'self' ${rule} https: data:; media-src 'self' ${rule} https: data:; script-src 'nonce-${nonce}' '${highlightjsInlineScriptHash}'; style-src 'self' ${rule} 'unsafe-inline' https: data:; font-src 'self' ${rule} https: data:;">`
+      default:
+      cspRule += `default-src 'none'; img-src 'self' ${rule} https: data:; media-src 'self' ${rule} https: data:; script-src 'nonce-${nonce}' '${highlightjsInlineScriptHash}'; style-src 'self' ${rule} 'unsafe-inline' https: data:; font-src 'self' ${rule} https: data:;`
+      break
   }
+     // Check if the enableKroki configuration is set to true
+  if (vscode.workspace.getConfiguration('asciidoc.extensions', null).get('enableKroki')) {
+    // Append the kroki-server-url to the CSP rule
+    cspRule += `; img-src 'self' ${kroki-server-url};`
```
Mogztter, line 83: are you sure it's working if we add the same rule twice? for instance, if the security is strict we will have `img-src` twice:

```
img-src 'self' ${rule} https: data:; img-src 'self' ${kroki-server-url};
```

Mogztter, line 83: `${kroki-server-url}` is undefined, you need to pass the value

haydencbarnes: Oof, I need to think about eliminating duplication of the getConfiguration code here too. Right now, I am thinking in the parser file:
```
// Check if the enableKroki configuration is set to true
  const enableKroki = vscode.workspace.getConfiguration('asciidoc.extensions', null).get('enableKroki');
  if (enableKroki) {
    // Check if the krokiServerUrl property is set
    if (vscode.workspace.getConfiguration('asciidoc.extensions', null).has('krokiServerUrl')) {
      // Get the krokiServerUrl property
      const krokiServerUrl = vscode.workspace.getConfiguration('asciidoc.extensions', null).get('krokiServerUrl');

      // Append the kroki-server-url to the CSP rule
      cspRule += `; img-src 'self' ${krokiServerUrl};`
    } 
```

And then change the if statement to: 

```
// Check if the krokiServerUrl property is set
if (vscode.workspace.getConfiguration('asciidoc.extensions', null).has('krokiServerUrl')) {
  // Get the krokiServerUrl property
  const krokiServerUrl = vscode.workspace.getConfiguration('asciidoc.extensions', null).get('krokiServerUrl');

  // Check if the enableKroki configuration is set to true
  const enableKroki = vscode.workspace.getConfiguration('asciidoc.extensions', null).get('enableKroki');
  if (enableKroki) {
    // Append the kroki-server-url to the CSP rule
    cspRule += `; img-src 'self' ${krokiServerUrl};`
  }
}
```

haydencbarnes: Will check tomorrow

Mogztter: `kroki-server-url` is an AsciiDoc attribute not a VS Code configuration, that's why you need to use the `AsciidocPreviewConfiguration` class.
The default value is `https://kroki.io` if the value is not explicitly defined.

haydencbarnes: Okay so I need tp add kroki-server-url as an AsciiDoc attribute to the AsciidocPreviewConfiguration class
```
export class AsciidocPreviewConfiguration {
  public static getForResource (resource: vscode.Uri) {
    return new AsciidocPreviewConfiguration(resource)
  }

  // ... other properties and methods

  public readonly krokiServerUrl: string
  
  private constructor (resource: vscode.Uri) {
    // ... other property assignments

    const asciidocConfig = vscode.workspace.getConfiguration('asciidoc', resource)
    this.krokiServerUrl = asciidocConfig.get<string>('preview.krokiServerUrl', 'https://kroki.io')
  }
}
```
Then the value of kroki-server-url should pass to the AsciidocPreviewConfiguration class. Then come back to the cspRules here and add:
```
// Check if the enableKroki configuration is set to true
const enableKroki = vscode.workspace.getConfiguration('asciidoc.extensions', null).get('enableKroki')
if (enableKroki) {
  // Get the value of the krokiServerUrl property from the AsciidocPreviewConfiguration class
  const config = AsciidocPreviewConfiguration.getForResource(resource)
  const krokiServerUrl = config.krokiServerUrl

  // Append the krokiServerUrl to the CSP rule
  cspRule += `; img-src 'self' ${krokiServerUrl};`
}
```

haydencbarnes: https://chrisguitarguy.com/2019/07/05/working-with-multiple-content-security-policy-headers/ - says that CSP headers can only get more restrictive, @Mogztter is this true in your experience?

- this article also shows that -https://w3c.github.io/webappsec-csp/#multiple-policies

haydencbarnes: If this is the case, I think I would need to write an if / else for the csp rules based on if the user has 'kroki-server-url' set not just enablekroki (because kroki.io is already allowed by default as you said.)

Mogztter: Not exactly, we need to load the AsciiDoc document and call `getAttribute()`: 

https://github.com/asciidoctor/asciidoctor-vscode/blob/117449b31216d9492e77d26b147610ed51bfee88/src/asciidocParser.ts#L203

You can have a look at: https://github.com/Mogztter/asciidoctor-kroki/blob/059510b500ad9e39733c27ba88b563736fc17ade/src/kroki-client.js#L76

Then, we need to pass this value to `AsciidoctorWebViewConverter` in order to compute the CSP correctly.

haydencbarnes: Do I need to load the Asciidoc document in the private constructor method or somewhere else within the 'AsciidocPreviewConfiguration' class? Confused about where to implement loading the document because I was initially under the impression that the AsciidocPreviewConfirguration class just pulled from the users vscode workspace? I can see that kroki-client you sent me now is doing something else on its end besides just using the users vscode workspace.

ggrossetie: I think we should load the Asciidoctor document using the `header_only` before we instantiate the `AsciidoctorWebViewConverter` class in:

https://github.com/asciidoctor/asciidoctor-vscode/blob/1e07c19864bc228661455f476a2d4e329c5fba45/src/asciidocParser.ts#L148-L157

```js
const asciidoctorDocument = processor.load(text, { header_only: true })
const krokiServerUrl = document.getAttribute('kroki-server-url') || 'https://kroki.io'

new AsciidoctorWebViewConverter(/* pass the krokiServerUrl here */)
```

As a side note, the AsciidoctorWebViewConverter constructor has quite a few arguments and we should probably refine it but we can do that later. For now, just add a new constructor argument named `krokiServerUrl`.

Does it make sense?

haydencbarnes: Okay, so to summarize before I make changes to my commit. I will be loading the Asciidoctor document using the `header_only` before we instantiate the `AsciidoctorWebViewConverter` in `asciidocParser.ts`, as shown by:

```
// load the asciidoc header only to get kroki-server-url attribute
const asciidoctorDocument = processor.load(text, { header_only: true })
const krokiServerUrl = document.getAttribute('kroki-server-url') || 'https://kroki.io'

const asciidoctorWebViewConverter = new AsciidoctorWebViewConverter( 
   doc, 
   editor, 
   cspArbiter.getSecurityLevelForResource(doc.uri), 
   cspArbiter.shouldDisableSecurityWarnings(), 
   this.contributionProvider.contributions, 
   previewConfigurationManager.loadAndCacheConfiguration(doc.uri), 
   antoraDocumentContext, 
   line,
   krokiServerUrl
 ) 
```
Then krokiServerUrl should be defined,
``

ggrossetie: Yes that's correct.

> Do I also need to add the krokiServerUrl here as initial data? -

I don't think so.

Issue #609, opened 2022-07-19
In the preview, a part of the table border is missing in "VS Code editor style" for a table with merged cells


ghost:
## Description

This table:

```adoc
[cols="1,1,1,1,1"]
[caption="Table {counter:table-number}: This is a Table", reftext="{table-caption} {table-number}", title="", id="some_table"]
|===
|*Column1*|*Column2*|*Column3*|*Column4*|*Column5*
.2+| Rows 1:2, Col 1
| Row 1, Col 2
| Row 1, Col 3
| Row1, Col 4
.2+| Rows 1:2, Col5
// No border generated
3+| Row 2, Col 2:4
|===
```

## System Information

chromeos: Version 103.0.5060.114 (Official Build) (64-bit)  also seen on mac osx
vscode: 1.69.2
asciidoctor-vscode: v2.9.8 and v3.0.1

## To Reproduce

Create file with above table and look at preview in vscode

### Failing screenshot

![Screenshot 2022-07-19 1 34 02 PM](https://user-images.githubusercontent.com/86068473/179844374-4c72b20d-4cba-4627-a26d-5c2eecbd7b11.png)

### Passing screenshot

![Screenshot 2022-07-19 1 31 56 PM](https://user-images.githubusercontent.com/86068473/179844448-7af218e1-d425-4513-8ae0-55bf6dd2428a.png)

haydencbarnes:
Looking into this

ggrossetie:
Sorry for the long silence on this one. I hadn't realized the table actually renders correctly with the default Asciidoctor stylesheet, and that the missing border was specific to the editor stylesheet (`asciidoc.preview.useEditorStyle`, on by default).

The editor stylesheet was still using Asciidoctor's legacy table border model (`border-collapse: separate` with each cell drawing only its right/bottom borders, plus `:last-child` rules stripping the outer edges). That model assumes the last cell of a row is also the visually last cell, which breaks with merged cells: in your example the `3+|` colspan cell is the last child of its `<tr>` (the two `.2+|` rowspan cells live in the previous row's `<tr>`), so its right border was wrongly removed and the segment next to the `Col5` cell disappeared.

I've switched the editor stylesheet to the same modern model the default stylesheet already uses (`border-collapse: collapse` with a full 1px border on every grid cell), so merged cells now keep complete borders.

PR #706, opened 2023-02-16
[#609]fix: adds border to last-child colspanned cells for better merged-cell styling

haydencbarnes:
/* git issue - https://github.com/asciidoctor/asciidoctor-vscode/issues/609
  If the last child cell was merged in the middle of three columns, then the middle cell of the three columns (but really last child cell) would not contain a right side border due to css "thinking" the cell was the most right cell and therefor be bordered by the frame element below this all. table.grid-cols>*>tr>.tableblock[colspan]:last-child sets a 1-pixel wide right border for the last cell in each row with a colspan attribute. This is needed because table.grid-all>*>tr>.tableblock:last-child removes the border from the last cell, but in the case of merged cells, the border is actually needed. Note: This may cause a small double border on the bottom most right cell of a colspanned row if it ends in the furthest right column.
*/

### DIFF
--- a/media/asciidoctor-editor.css
+++ b/media/asciidoctor-editor.css
@@ -1672,6 +1672,18 @@ table.grid-all>*>tr>.tableblock:last-child,
 table.grid-cols>*>tr>.tableblock:last-child {
   border-right-width: 0
 }
+ /* git issue - https://github.com/asciidoctor/asciidoctor-vscode/issues/609
+  If the last child cell was merged in the middle of three columns, then the middle cell of the three columns (but really last child cell)
+  would not contain a right side border due to css "thinking" the cell was the most right cell and therefor be bordered by
+  the frame element below this all.
+  table.grid-cols>*>tr>.tableblock[colspan]:last-child (below) sets a 1-pixel wide right border for the last cell in each row with a
+  colspan attribute. This is needed because table.grid-all>*>tr>.tableblock:last-child removes the border from the last cell,
+  but in the case of merged cells the border is actually needed. Note: This may cause a small double border on the bottom
+  most right cell of a colspanned row if it ends in the furthest right column.
+*/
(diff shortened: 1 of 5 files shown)


haydencbarnes:
@ggrossetie looks like refactoring like in https://github.com/asciidoctor/asciidoctor-vscode/blob/fba9fb79e63c09a09e306e352ee294d8b94190d0/media/asciidoctor-default.css#L1402-L1440 

actually might take care of everything we need, was there any reason the editor style targets the last child here when we could just have the border cover all the cells the same as shown in the default stylesheet?
```
table.grid-cols>*>tr>.tableblock:last-child {
  border-right-width: 0
}
```
I think the issue with my commit rn is that it may introduce a double border on the bottom most right cell of a colspanned row if it ends in the furthest right column.

haydencbarnes:
Still wanting to make progress on this, will try refactor later this week

ggrossetie:
@haydencbarnes I believe it's an impossible problem to solve: 

> The default CSS does not handle rowspans. That's a limitation of having generic CSS. To fix it, I recommend adding a role to the table and customizing the CSS for that table. There's just no way that I can see that we could possibly address all the rowspan permutations. (The same goes for certain cases of colspans).

https://github.com/asciidoctor/asciidoctor/issues/3447

> Otherwise, the only way to solve this problem is for the converter to add hints to the cells so we can detect when a cell extends into another column or row and touches the boundary of the table.

If we aren't using the latest version of [asciidoctor.css](https://github.com/asciidoctor/asciidoctor/blob/main/src/stylesheets/asciidoctor.css) you can try to upgrade to see if it improves things. Otherwise, and as mentioned by Dan, the only way to solve this problem is to add hints (i.e., CSS classes) to the cells.

haydencbarnes:
@ggrossetie do you mean something like this - https://github.com/fomantic/Fomantic-UI/pull/1334 ?

PR #721, opened 2023-06-16
WIP Search in attribute value during completion

### LINKED ISSUE #720: Search in attribute value during completion
Currently, for attributes, when using completion the elements are filtered based on the attribute key. it would be nice to also search with the attribute value.

use case is coming from our Customer Content Service, they have 2 cases where it would be very useful:
- (re-)naming of Software/product lines: attributes are shared across a lot of documentation files. The writers usually knows a part of the product name, which can be very long and not all parts of it are part of the key. After creation, it can be very complicated to rename the key. 
- some attributes are generated and the key is a random generated number

I feel like extrapolating from these use cases, it would be nice in more cases to have the search in the values of the attributes too.

https://github.com/asciidoctor/asciidoctor-vscode/assets/1105127/7a3cda10-abfa-461e-8a4d-6f1ef0de0681

https://github.com/asciidoctor/asciidoctor-vscode/assets/1105127/1c0a37ee-15ab-40fc-9495-778d342b5c4d

apupier:
fixes #720

**IMPORTANT: Please do not create a pull request without creating an issue first.**

*Any change needs to be discussed before proceeding. Failure to do so may result in the rejection of the pull request.*

Put `resolves #XXX` in your comment to auto-close the issue that your pull request fixes.

todo:
- provide tests


ggrossetie:
Looks good, thanks 👍🏻

Issue #719, opened 2023-06-03
folding of properties/attributes


maxandersen:
would be great to collapse properties/attributes to remove "metadata noise".

Example from using asciidoctor reveal.js you have this block that you don't really need to see for normal editing:

```
[.blackback]
= Embracing a new thing :date: 2022-05-10
:revealjs_theme: white
:customcss: css/custom.css
:favicon: ./images/favicon.ico
// better 
:title-slide-background-image: images/blackspace.png
:slide-background-image: images/slide2-white.png
:revealjs_fragmentInURL: true
:revealjs_hash: true
:revealjs_controls: true
:revealjs_controlsLayout: edges
:revealjs_controlsTutorial: true
:revealjs_slideNumber: c/t
:revealjs_showSlideNumber: speaker
:revealjs_autoPlayMedia: true
:revealjs_totalTime: 2700
//:revealjs_parallaxBackgroundImage:  images/mntbackground.jpg
//:revealjs_parallaxBackgroundSize: 4936px 2092px
:source-highlighter: highlight.js
// leave empty or pick style from gallery at https://highlightjs.org/static/demo/ and find url at https://cdnjs.com/libraries/highlight.js
:highlightjs-theme: https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.5.1/styles/base16/solarized-dark.min.css
:revealjs_preloadIframes: true
:icons: font

Main text
```

would be nice to have that be:

```
   [.blackback]
   = Embracing Quarkus 3
> (attributes)

Main text
```

ggrossetie:
That's an excellent idea! 👍🏻

PR #723, opened 2023-06-16
Provide folding range for sibling attributes

apupier:
part of #719

basic use case covered, other cases potentially to cover:
- include comment lines
- include emtpy lines

https://github.com/asciidoctor/asciidoctor-vscode/assets/1105127/44727ea4-356d-4db9-9882-120d121fe157

### DIFF
--- a/src/test/foldingProvider.test.ts
+++ b/src/test/foldingProvider.test.ts
@@ -394,6 +394,24 @@ this is the same paragraph`)
       ])
     })
   })
+
+  suite('getMultiAttributesFoldingRanges', () => {
+    test('Should fold on a group of attributes ', () => {
+      const folds = getFoldsForDocument(
+        `this is a paragraph
+
+:attribute1: value 1
+:attribute2: value 2
+:attribute3: value 3
+:attribute4: value 4
+
+this is a paragraph`)
+      assert.strictEqual(folds.length, 1, 'expecting 1 fold')
+      assert.deepStrictEqual(folds, [
+        new vscode.FoldingRange(2, 5),
+      ])
+    })
+  })
 })
 
 function getFoldsForDocument (contents: string) {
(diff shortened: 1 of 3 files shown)


src/test/foldingProvider.test.ts
```diff
@@ -394,6 +394,23 @@ this is the same paragraph`)
       ])
     })
   })
+
+  suite('getMultiAttributesFoldingRanges', () => {
+    test.only('Should fold on a group of attributes ', () => {
```
apupier, line 399: .only to remove

ggrossetie:
I think the only uncovered use case is a multi-line attribute:

```adoc
= Document
:description: If you have a very long line of text \
that you need to substitute regularly in a document, \
you may find it easier to split the value neatly in the header \
so it remains readable to folks looking at the AsciiDoc source.
:attr2: value
:attr3: value

content
```

ggrossetie:
### 1. include comment lines

I don't know about that. I think a comment is an indication that attributes are "grouped":

```adoc
= Document
// urls
:url-foo: foo
:url-bar: bar
// built-in
:toc:
:author: Aurélien

content
```

We might want to fold all document attributes at once but I think we should also allow to fold each "group".

### 2. include empty lines

It's important to note that:

> The document header may not contain empty lines. The first empty line the processor encounters after the document header begins marks the [end of the document header](https://docs.asciidoctor.org/asciidoc/latest/document/header/#when-does-the-document-header-end) and the start of the document body.

https://docs.asciidoctor.org/asciidoc/latest/document/header/#document-header-structure

It might be misleading if we fold all attributes in the document header:

```
= Document
:doc-attr1: value
:doc-attr2: value

:not-doc-attr: value

content
```

For attributes defined in the body I think it's fine:

```
= Document
:attr1: value
:attr2: value

content

:body-attr1: value
:body-attr2: value

:body-attr3: value
:body-attr4: value

more content
```

But again it makes sense to fold per "group". I think that's the whole purpose of adding an empty line, to visually create two groups of attributes.


Anyway, we can discuss that in follow-up issues.

PR #733, opened 2023-06-23
Folding mix of Attributes and single-line comments

### LINKED ISSUE #719: folding of properties/attributes

apupier:
It removes the possibility to have Folding range for the first group of attributes or first group of single-line comments. This is due to a limitation of VS Code which doesn't handle several folding ranges starting on the same line. This limitation lead me to add some non negligeable code complexity.

resolves #719

https://github.com/asciidoctor/asciidoctor-vscode/assets/1105127/5649fe55-47de-4265-b15c-c3d0047c6e6b

### DIFF
--- a/src/test/foldingProvider.test.ts
+++ b/src/test/foldingProvider.test.ts
@@ -413,6 +413,77 @@ this is a paragraph`)
       ])
     })
   })
+
+  suite('getMultiAttributeAndSingleLineCommentsFoldingRanges', () => {
+    test('Should fold on a group of mix of attributes and single line comments ', () => {
+      const folds = getFoldsForDocument(
+        `this is a paragraph
+
+// A single-line comment.
+:an-attribute: value of attribute
+// A second single-line comment.
+:a-second-attribute: value of second attribute
+
+this is a paragraph`)
+      assert.strictEqual(folds.length, 1, 'expecting 1 fold')
+      assert.deepStrictEqual(folds, [
+        new vscode.FoldingRange(2, 5),
+      ])
+    })
+    test('Should fold on a group of mix of attributes and single line comments starting with 2 attributes', () => {
+      const folds = getFoldsForDocument(
+        `this is a paragraph
+
+:an-attribute1: value of attribute
+:an-attribute2: value of attribute
+// A single-line comment.
+:an-attribute: value of attribute
+// A second single-line comment.
+:a-second-attribute: value of second attribute
+
+this is a paragraph`)
+      assert.strictEqual(folds.length, 1, 'expecting 1 fold')
+      assert.deepStrictEqual(folds, [
+        new vscode.FoldingRange(2, 7),
+      ])
+    })
+    test('Should fold on a group of mix of attributes and single line comments starting with 2 line comments', () => {
+      const folds = getFoldsForDocument(
+        `this is a paragraph
+
+// A single-line comment.
+// A second single-line comment.
+:an-attribute: value of attribute
+// A third line comment
+:a-second-attribute: value of second attribute
+
+this is a paragraph`)
+      assert.strictEqual(folds.length, 1, 'expecting 1 fold')
+      assert.deepStrictEqual(folds, [
+        new vscode.FoldingRange(2, 6),
+      ])
+    })
+    test('Should fold on a group of mix of attributes and single line comments and include next groups', () => {
+      const folds = getFoldsForDocument(
+        `this is a paragraph
+
+// A single-line comment.
+// A second single-line comment.
+:an-attribute: value of attribute
+:another-attribute:
+// A third line comment
+// as a potential folded group
+:a-third-attribute: value of third attribute
+
+this is a paragraph`)
+      assert.strictEqual(folds.length, 3, 'expecting 3 folds')
+      assert.deepStrictEqual(folds, [
+        new vscode.FoldingRange(4, 5),
+        new vscode.FoldingRange(6, 7, vscode.FoldingRangeKind.Comment),
+        new vscode.FoldingRange(2, 8),
+      ])
+    })
+  })
 })
 
 function getFoldsForDocument (contents: string) {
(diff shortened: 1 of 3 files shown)


src/test/foldingProvider.test.ts
```diff
@@ -11,7 +11,7 @@ import { InMemoryDocument } from './inMemoryDocument'
 
 const testFileName = vscode.Uri.file('test.adoc')
 
-suite('asciidoc.FoldingProvider', () => {
+suite.only('asciidoc.FoldingProvider', () => {
```
apupier, line 14: oops .only, will repush

apupier:
> Arguably, we could collapse the document header when we click on the document title. The downside is that you cannot collapse the entire file but is that really useful?

I do not understand why it would interfere with collapsing the whole file.

I feel like it would be to implement the folding for each section delimited by `=` and the various levels of `==...=`.

ggrossetie:
I believe that, currently, if you fold on the document title, it will fold the whole document, correct?

What I'm proposing is that if you fold on the document title, it will only fold the document header (i.e. document attributes, author line, etc...).

By doing that, my guess is that you can simplify the logic for attributes folding? You can fold a group of attributes until a comment or an empty line is found.

apupier:
> I believe that, currently, if you fold on the document title, it will fold the whole document, correct?

https://github.com/asciidoctor/asciidoctor-vscode/assets/1105127/c894a12b-ce96-4421-b238-81d50874bfc5


no, there is no fold on the document title

apupier:
> What I'm proposing is that if you fold on the document title, it will only fold the document header (i.e. document attributes, author line, etc...).
> 
> By doing that, my guess is that you can simplify the logic for attributes folding? You can fold a group of attributes until a comment or an empty line is found.

So implementing it this way will  effectively simplify code but I think having the fold of a whole "titled section" is more interesting.

Another side thoughts, if a folding section is missing, end users are nto block and can create their own custom folding section https://code.visualstudio.com/docs/editor/codebasics#_fold-selection

ggrossetie:
> but I think having the fold of a whole "titled section" is more interesting.

What do you mean? My idea was to have a special fold on the document title (single `=`) but keep the existing behavior for section titles (two or more `=`).

> Another side thoughts, if a folding section is missing, end users are nto block and can create their own custom folding section https://code.visualstudio.com/docs/editor/codebasics#_fold-selection

Oh I didn't know that! We should probably mention that in the documentation!

ggrossetie:
I think I understand what's bothering you with my proposal.
In the following case, we won't be able to fold the whole block of attributes/comments since my proposal is basically to fold until a comment is found.

```
this is a paragraph

:an-attribute1: value of attribute
:an-attribute2: value of attribute
// A single-line comment.
:an-attribute: value of attribute
// A second single-line comment.
:a-second-attribute: value of second attribute

this is another paragraph.
```

Having said that, I don't think this is a common use case. The document header is where we usually have a mix of comments and attributes. Attributes defined in the body of the document are "rare". It's also extremely uncommon to have a block of comments/attributes in the body.

So I think we should fold attributes until we found a comment or something that's not an attribute.

Does it make sense?

ggrossetie:
I think it's difficult to decide because it depends 😄 

In the example below, we most likely want to fold all the attribute entries + comment:
```
this is a paragraph

:an-attribute1: value of attribute
:an-attribute2: value of attribute
//:an-attribute3: value of attribute
:an-attribute4: value of attribute

this is a paragraph
```

However in the following example I would argue that we want to fold per "group":

```
= Embracing a new thing
// localization
:example-caption: Exemple
:tip-caption: Astuce
// reveal.js options
:revealjs_fragmentInURL: true
:revealjs_hash: true
:revealjs_controls: true
:revealjs_controlsLayout: edges
:revealjs_controlsTutorial: true
:revealjs_slideNumber: c/t
:revealjs_showSlideNumber: speaker
:revealjs_autoPlayMedia: true
:revealjs_totalTime: 2700
// style
:icons: font
:source-highlighter: highlight.js

content...
```

In other words, comments are used as headers/delimiters so it makes sense to fold the localization group, the reveal.js options group and the style group independently. 

For reference, the Intellij IDEA extension does not seem to provide this feature.

@mojavelinux @ahus1 I would love your input on this issue 🥺

mojavelinux:
My intuition says that groups of attribute entries would be folded behind a leading line comment (the line comment being the fold line). (I'm not sure if that's feasible to implement, but it would make sense to me as a user).

ggrossetie:
> My intuition says that groups of attribute entries would be folded behind a leading line comment (the line comment being the fold line).

I agree 👍🏻

apupier:
After 2 years and my role changing at work, I must face that I won't have time to work again on this issue. I'm closing the PR. Feel free to reuse (or not) the part that you need.

Issue #727, opened 2023-06-19
Attribute coming from include files are missing in completion with 3.0.x


apupier:
using a file like my.adoc:
```
== test

include::ref1.adoc[]

```
and a file ref1.adoc:
```
:aref1: test
```

calling completion in first file does not always provide the attributes of the included file where other attributes are defined.

one case working:
![Screenshot from 2023-06-12 15-21-54](https://github.com/asciidoctor/asciidoctor-vscode/assets/1105127/294c0fdd-7624-4183-90e4-6e692099101e)

one case not working:
![Screenshot from 2023-06-19 17-25-53](https://github.com/asciidoctor/asciidoctor-vscode/assets/1105127/a50998be-ee83-487b-a931-3c8840815009)

apupier:
it works fine with 2.9.8

apupier:
I tried to add several additional parameters from the [ProcessorOptions](https://github.com/asciidoctor/asciidoctor.js/blob/cba1da0a23f79b14be3a37b99a91f0e1b4cef0f3/packages/core/types/index.d.ts#L622):
https://github.com/apupier/asciidoctor-vscode/blob/261224c2f3f9f27bcb0fc7c03d0c860d7bb2c653/src/asciidocParser.ts#L120

```
    const document = processor.load(textDocument.getText(), {
      attributes: {
        'allow-uri-read': '',
        'env-vscode': '',
        env: 'vscode',
      },
      extension_registry: registry,
      sourcemap: true,
      safe: 'unsafe',
      'allow-uri-read': '',
      parse: true,
      catalog_assets: true,
      ...(baseDir && { base_dir: baseDir }),
    })
 ```

without success

ggrossetie:
We only retrieve document attributes using: https://github.com/apupier/asciidoctor-vscode/blob/eece7d6dbcb7fcf3596cf3efa8168535d298990a/src/features/attributeReferenceProvider.ts#L10-L11

Attributes defined in the body are processed during the parse phase. Previously, the implementation was using a regular expression to get any attributes (i.e, any line that starts with `:`).

If you remove the empty lines it will work because all attributes will be defined as document attributes.

ggrossetie:
The logic is here:

https://github.com/asciidoctor/asciidoctor-vscode/blob/45302b52a57c8e3ffdd37cdca3a4f7145df7be23/src/features/attributeReferenceProvider.ts#L7-L8

apupier:
[...]

PR #734, opened 2023-06-23
Fix Attributes defined in another file which is included

apupier:
resolves of #727

image

### DIFF
--- a/src/asciidocParser.ts
+++ b/src/asciidocParser.ts
@@ -76,17 +76,13 @@ export class AsciidocParser {
     highlightjsBuiltInSyntaxHighlighter.$register_for('highlight.js', 'highlightjs')
     const baseDir = AsciidocTextDocument.fromTextDocument(textDocument).getBaseDir()
     const templateDirs = this.getTemplateDirs()
-    const options: { [key: string]: any } = {
-      attributes: {
-        'env-vscode': '',
-        env: 'vscode',
-        ...asciidoctorAttributes,
-      },
-      backend,
-      extension_registry: registry,
-      header_footer: true,
-      safe: 'unsafe',
-      ...(baseDir && { base_dir: baseDir }),
+    const options: { [key: string]: any } = AsciidocParser.getDefaultAsciidoctorOptions(baseDir)
+    options.extension_registry = registry
+    options.header_footer = true
+    options.attributes = {
+      'env-vscode': '',
+      env: 'vscode',
+      ...asciidoctorAttributes,
     }
     if (templateDirs.length !== 0) {
       options.template_dirs = templateDirs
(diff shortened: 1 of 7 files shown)


src/asciidocParser.ts
```diff
@@ -110,7 +110,7 @@ export class AsciidocParser {
 
   // Load
 
-  public static load (textDocument: SkinnyTextDocument): { document: Asciidoctor.Document, baseDocumentIncludeItems: IncludeItems } {
+  public static loadWithoutParsingIncludes (textDocument: SkinnyTextDocument): { document: Asciidoctor.Document, baseDocumentIncludeItems: IncludeItems } {
```
ggrossetie, line 113: Since we don't use `document`, I think we could rename this function to `getBaseDocumentIncludeItems`.

src/asciidocParser.ts
```diff
@@ -134,6 +134,23 @@ export class AsciidocParser {
     }
   }
 
+  public static load (textDocument: SkinnyTextDocument): Asciidoctor.Document {
+    const memoryLogger = processor.MemoryLogger.create()
+    processor.LoggerManager.setLogger(memoryLogger)
+    asciidoctorFindIncludeProcessor.resetIncludes()
```
ggrossetie, line 140: I don't think we need to call `resetIncludes` here.

src/asciidocParser.ts
```diff
@@ -134,6 +134,23 @@ export class AsciidocParser {
...
+    return processor.load(textDocument.getText(), {
+      attributes: {
+        'env-vscode': '',
+        env: 'vscode',
+      },
+      sourcemap: true,
+      safe: 'unsafe',
+      parse: true,
+      ...(baseDir && { base_dir: baseDir }),
+    })
```
ggrossetie, line 151: It might be worth to declare a function that build the "default" Asciidoctor options:


```js
const options = getDefaultAsciidoctorOptions(baseDir)
```

src/asciidocParser.ts
ggrossetie: Never mind, we are actually using `document.getSourceLines()` 😬 

Having said that, I still think that we should rename this function.
Also, I wonder if `document.getSourceLines()` is actually the same as `textDocument.getText().split('\n')`. If that's the case then we don't need to return `document`.

src/test/attributeRefCompletionProvider.test.ts
```diff
@@ -59,4 +59,30 @@ somethingVeryDifferent`))
     const completionsItems = new AttributeReferenceProvider().provideCompletionItems(file, new Position(1, 22))
     assert.notStrictEqual(completionsItems.length, 0, 'There are Completion provided although none are expected.')
   })
+  test('Should return attribute key defined in another file', async () => {
```
ggrossetie, line 62: I love to see new tests! Thanks for adding them 🥇

src/asciidocParser.ts
apupier: used `textDocument.getText().split('\n')` and tests are passing

src/asciidocParser.ts
```diff
@@ -76,17 +76,13 @@ export class AsciidocParser {
...
-      header_footer: true,
-      safe: 'unsafe',
-      ...(baseDir && { base_dir: baseDir }),
+    const options: { [key: string]: any } = AsciidocParser.getDefaultAsciidoctorOptions(baseDir)
+    options.extension_registry = registry
+    options.header_footer = true
+    options.attributes = {
+      'env-vscode': '',
+      env: 'vscode',
+      ...asciidoctorAttributes,
```
apupier, line 85: maybe possible to have a better TypeScript notation here

src/test/attributeRefCompletionProvider.test.ts
```diff
@@ -18,7 +18,7 @@ suite('Attribute ref CompletionsProvider', () => {
     createdFiles = []
   })
   test('Should return attribute key defined in same file', async () => {
-    const fileToAutoComplete = vscode.Uri.file(`${root}/fileToAutoComplete-attributeRef.adoc`)
+    const fileToAutoComplete = vscode.Uri.file(`${root}/fileToAutoComplete-attributeRef-samfile.adoc`)
```
ggrossetie, line 21:
```suggestion
    const fileToAutoComplete = vscode.Uri.file(`${root}/fileToAutoComplete-attributeRef-samefile.adoc`)
```

ggrossetie:
Looks good to me, thanks!

PR #737, opened 2023-06-29
WIP - attempt to support remote includes

### LINKED ISSUE #731: Support remote includes
Currently remote includes are not resolved (neither in the preview, neither in completion of attributes)

https://docs.asciidoctor.org/asciidoc/latest/directives/include-uri/#reference-include-content-by-uri

example:
```
include::https://gist.githubusercontent.com/apupier/528a635931932adbdf755303e0116314/raw/f55d188f903ede2a9f3a25c7deeb28046810fe5c/test-remote-attribute.adoc[]

== test

{remote-test}
```

preview gives:
image

apupier:
from what i understood, placing the allow-uri-read in attributes should have done the thing but it is not working... to investigate

resolves #731

### DIFF
--- a/src/asciidocParser.ts
+++ b/src/asciidocParser.ts
@@ -80,6 +80,7 @@ export class AsciidocParser {
     options.extension_registry = registry
     options.header_footer = true
     options.attributes = {
+      'allow-uri-read': '',
       'env-vscode': '',
       env: 'vscode',
       ...asciidoctorAttributes,
@@ -130,6 +131,7 @@ export class AsciidocParser {
   private static getDefaultAsciidoctorOptions (baseDir: string): any {
     return {
       attributes: {
+        'allow-uri-read': '',
         'env-vscode': '',
         env: 'vscode',
       },
@@ -204,6 +206,7 @@ export class AsciidocParser {
     attributes['env-vscode'] = ''
     attributes.env = 'vscode'
     attributes['relfilesuffix@'] = '.adoc'
+    attributes['allow-uri-read'] = ''
 
     const antoraSupport = await AntoraSupportManager.getInstance(context.workspaceState)
     const antoraAttributes = await antoraSupport.getAttributes(doc.uri)


apupier:
@ggrossetie any hints how to investigate on this one?

ggrossetie:
The underlying error is:

```
Error: Command failed: "/usr/share/code/code" "/home/guillaume/Workspace/opensource/asciidoctor-vscode/node_modules/unxhr/lib/request.js" --ssl="true" --encoding="utf8" --request-options="{\"host\":\"raw.githubusercontent.com\",\"port\":443,\"path\":\"/asciidoctor/asciidoctor-vscode/master/CONTRIBUTING.adoc\",\"method\":\"GET\",\"headers\":{\"User-Agent\":\"node-XMLHttpRequest\",\"Accept\":\"*/*\",\"Host\":\"raw.githubusercontent.com\"},\"agent\":false,\"withCredentials\":false}"
````

We are using `unxhr` to fetch remote content synchronously (because we must resolve include synchronously). To do that, we use `execSync`.

I believe that the issue is that we are using `${process.execPath}` which in this case resolves to `/usr/share/code/code` (which seems to be unable to execute a Node script).

https://github.com/ggrossetie/unxhr/blob/4fbc39978b99180c24cd31ac802c2b4ae5dca3c6/lib/XMLHttpRequest.js#L520

Not sure how we can execute a Node script using `code` (since we cannot be sure that `node` will be available in the `PATH`).

apupier:
> I believe that the issue is that we are using `${process.execPath}` which in this case resolves to `/usr/share/code/code` (which seems to be unable to execute a Node script).

seems we are not the only ones to hit this kind of problem: https://stackoverflow.com/questions/75103016/node-execpath-references-vscode-installation-instead-of-path-to-node-installatio

apupier:
I'm wondering if we can use child_process.execFileSync https://nodejs.org/api/child_process.html#child_processexecfilesyncfile-args-options instead of child_process.execSync in unxhr

apupier:
something to try is to use ELECTORON_RUN_AS_NODE https://www.electronjs.org/docs/latest/api/environment-variables#electron_run_as_node

https://github.com/microsoft/vscode-discussions/discussions/746#discussioncomment-6719320

apupier:
After 2 years and my role changing at work, I must face that I won't have time to work again on this issue. I'm closing the PR. Feel free to reuse (or not) the part that you need.

PR #739, opened 2023-06-30
Support remote includes for PDF export

apupier:
part of #731

https://github.com/asciidoctor/asciidoctor-vscode/assets/1105127/a7bf6846-6d7d-4410-afe7-a7f113c2129f

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Improvements
 
 - provide folding for list of sibling attributes by @apupier (#719)
+- support remote includes when exporting in PDF without additional setting configuration (#731)
 
 ### Bug fixes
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Improvements
 
 - provide folding for list of sibling attributes by @apupier (#719)
+- support remote includes when exporting in PDF (#731)
```
ggrossetie, line 8: Since we can add additional command line arguments, I guess we should say: "by default, allow remote includes when exporting in PDF"?

apupier: with current implementation, it is not only by default. The parameter is always provided. The parameter from the settings are appended, they are not replacing some parameters that are always passed.

ggrossetie: Sorry I wasn't clear.
My point was that we already support remote includes when exporting in PDF since we can set `-a allow-uri-read` in the extension settings. I was suggesting to reword the changelog to mention that remote includes are now enabled by default when exporting to PDF.

PR #768, opened 2023-07-25
fix regexes in #quote-paragraph rule, and downstream

### LINKED ISSUE #767: Grammar rule won't match block-attribute-list that contains an inline macro
There is an unaddressed problem in the TextMate grammar that was brought over from Atom in Feb. 2020 (see [file history for the grammar-file, Asciidoctor.json)](https://github.com/asciidoctor/asciidoctor-vscode/commits/master/syntaxes/Asciidoctor.json)).

As it stands, the regex in the #quote-paragraph rule will not match any block-attribute-list that contains, itself, an inline macro -- such as a URL macro or a bibliographic-citation macro. Downstream from that rule, there are weaknesses also in the #block-attribute-inner rule. The regex-fixes I'm going to show in a PR here (in just a minute) are the upshot of an earlier discussion of the whole matter, still viewable in [Atom PR 197](https://github.com/asciidoctor/atom-language-asciidoc/pull/197) -- where also there are screen-shots and syntax-examples; please see this earlier discussion for details. 

(Sadly, that discussion was started almost 2 years **after** the grammar was ported here from Atom... Too bad I was barking up the wrong tree... Atom instead of VS Code! :-|)

This problem remains as of v. 3.1.3. Implementing them in that version, I've tested these fixes that I'm going to submit in my PR. The system information is:

Version: 1.79.2 (system setup)
Commit: 695af097c7bd098fbf017ce3ac85e09bbc5dda06
Date: 2023-06-14T08:57:04.379Z
Electron: 22.5.7
Chromium: 108.0.5359.215
Node.js: 16.17.1
V8: 10.8.168.25-electron.0
OS: Windows_NT x64 10.0.19045

andrewcarver:
Addresses a problem in the grammar: As it stands, the regex in rule #quote-paragraph will not match any block-attribute-list that contains, itself, an inline macro -- such as a URL macro or a bibliographic-citation macro. Downstream from that rule, there are weaknesses also in the #block-attribute-inner rule. The fixes I've applied here are based on earlier discussion of this whole matter in [atom-language-asciidoc PR 197](https://github.com/asciidoctor/atom-language-asciidoc/pull/197) -- where there are screen-shots and syntax-examples.

Resolves #767.

### DIFF
--- a/syntaxes/Asciidoctor.json
+++ b/syntaxes/Asciidoctor.json
@@ -337,7 +337,7 @@
         {
           "comment": "attributes",
           "name": "markup.meta.attribute-list.asciidoc",
-          "match": "(?<=\\{|,|.|#|\"|'|%)([^\\],.#%]+)",
+          "match": "(?<=\\{|,|\\.|#|\"|'|%)(?:[^\\],.#%]|\\](?=[ \\t]*\\S))+",
           "captures": {
             "0": {
               "patterns": [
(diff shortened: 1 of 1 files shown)


ggrossetie:
The grammar has been significantly modified since then. As mentioned in the issue, it is important to provide an example so that we can add tests for the grammar.

PR #845, opened 2024-02-04
add doc role when the asciidoc page is used in an antora context

### LINKED ISSUE #844: Improve antora experience by putting a doc div into generated preview html
I am currently looking into a way to improve the antora experience for our writers by providing a custom stylesheet that is as near as possible to our antora ui.

While looking into it I found the configuration to set the preview stylesheet. The problem now is that the default antora stylesheet are expecting everything to be under `.doc` div but in the preview html no `.doc` div is available. Everything is under an article.

**Request**
* when antora support is set to true
* generate the preview html under a div with the class="doc".

**Benefits**
* It would be possible to put the whole antora ui-bundle/_css/ under e.g. .asciidoctor/preview and set the `asciidoc.preview.style` to ./asciidoctor/preview/css/site.css
* Easy to get the same ui in the preview as in the antora page

lask79:
Resolves #844

### DIFF
--- a/src/asciidoctorWebViewConverter.ts
+++ b/src/asciidoctorWebViewConverter.ts
@@ -189,7 +189,7 @@ export class AsciidoctorWebViewConverter {
       <body${node.getId() ? ` id="${node.getId()}"` : ''} class="${this.getBodyCssClasses(node)}">
         ${headerDocinfo}
         ${this.getDocumentHeader(node)}
-        <div id="content">
+        <div id="content" ${this.antoraDocumentContext ? 'class="doc"' : ''}>
           ${node.getContent()}
         </div>
         ${this.generateFootnotes(node)}
--- a/src/asciidoctorWebViewConverter.ts
+++ b/src/asciidoctorWebViewConverter.ts
@@ -189,7 +189,7 @@ export class AsciidoctorWebViewConverter {
       <body${node.getId() ? ` id="${node.getId()}"` : ''} class="${this.getBodyCssClasses(node)}">
         ${headerDocinfo}
         ${this.getDocumentHeader(node)}
-        <div id="content" ${this.antoraDocumentContext ? 'class="doc"' : ''}>
+        <div id="content"${this.antoraDocumentContext ? ' class="doc"' : ''}>
           ${node.getContent()}
         </div>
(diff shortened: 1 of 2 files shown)


src/asciidoctorWebViewConverter.ts
```diff
@@ -189,7 +189,7 @@ export class AsciidoctorWebViewConverter {
       <body${node.getId() ? ` id="${node.getId()}"` : ''} class="${this.getBodyCssClasses(node)}">
         ${headerDocinfo}
         ${this.getDocumentHeader(node)}
-        <div id="content">
+        <div id="content" ${this.antoraDocumentContext ? 'class="doc"' : ''}>
```
ggrossetie, line 192: Avoid an extra space when `this.antoraDocumentContext` is undefined:

```suggestion
        <div id="content"${this.antoraDocumentContext ? ' class="doc"' : ''}>
```

ggrossetie:
Would you be able to add a test case in https://github.com/asciidoctor/asciidoctor-vscode/blob/master/src/test/asciidoctorWebViewConverter.test.ts to test with and without an `antoraDocumentContext`?

lask79:
Sure I will try to do that

lask79:
I added testcases but I cannot test them locally so please run the test and tell me if there needs to be changed something. 

I am getting the following error when I run the tests (independent from my change):
Exit code:   SIGTRAP

I am currently woring on a WSL2 - Ubuntu Image with VSCode Remote Development. Maybe that is related.

lask79:
Ok, I am now able to run the Tests using the debugger. Now I have to figure out to test the content div. I will take a look later.

ggrossetie:
The reason is that we are using `const html = processor.convert(input, { converter: asciidoctorWebViewConverter })`. In this case, Asciidoctor will convert inline markup: https://docs.asciidoctor.org/asciidoctor/latest/api/convert-strings/#convert-inline-markup-only

You should create a new test case to convert a standalone document: https://docs.asciidoctor.org/asciidoctor/latest/api/convert-strings/#standalone-output

lask79:
fixed tests

lask79:
Is anything missing from my side to merge this?

ggrossetie:
No all good 👍🏻

PR #853, opened 2024-03-11
Fix non working xref in preview pane

### LINKED ISSUE #425: Does asciidoc have issues with xref and include?

ViToni:
Both `xref:` and `link:` create `<a>` elements. For `xref:` the `data-href` attribute was missing so WebView didn't treat it as a regular link. Now `xref:` and `link:` are created the same way for the WebView thus allowing `xref:` navigation.

Resolves #425.

### DIFF
--- a/preview-src/index.ts
+++ b/preview-src/index.ts
@@ -82,7 +82,7 @@ const onUpdateView = (() => {
 const updateImageSizes = throttle(() => {
   const imageInfo: { id: string, height: number, width: number }[] = []
   const images = document.getElementsByTagName('img')
-  if (images) {
+  if (images && images.length > 0) {
     let i
     for (i = 0; i < images.length; i++) {
       const img = images[i]
--- a/src/asciidoctorWebViewConverter.ts
+++ b/src/asciidoctorWebViewConverter.ts
@@ -202,12 +202,56 @@ export class AsciidoctorWebViewConverter {
       </body>
       </html>`
     }
-    if (nodeName === 'inline_anchor' && node.type === 'link') {
-      const href = isSchemeBlacklisted(node.target) ? '#' : node.target
-      const id = node.hasAttribute('id') ? ` id="${node.id}"` : ''
-      const role = node.hasAttribute('role') ? ` class="${node.getRole()}"` : ''
-      const title = node.hasAttribute('title') ? ` title="${node.title}"` : ''
-      return `<a href="${href}"${id}${role}${title} data-href="${href}">${node.text}</a>`
+    if (nodeName === 'inline_anchor') {
+      if (node.type === 'link') {
+        const href = isSchemeBlacklisted(node.target) ? '#' : node.target
+        const id = node.hasAttribute('id') ? ` id="${node.id}"` : ''
+        const role = node.hasAttribute('role') ? ` class="${node.getRole()}"` : ''
+        const title = node.hasAttribute('title') ? ` title="${node.title}"` : ''
+        return `<a href="${href}"${id}${role}${title} data-href="${href}">${node.text}</a>`
+      }
+      if (node.type === 'xref') {
+        const attrs = []
+        attrs.push(` href="${node.target}"`)
+
+        if (node.hasAttribute('id')) { attrs.push(` id="${node.id}"`) }
+        if (node.hasAttribute('role')) { attrs.push(` class="${node.getRole()}"`) }
+        if (node.hasAttribute('title')) { attrs.push(` title="${node.title}"`) }
+
+        attrs.push(` data-href="${node.target}"`)
+
+        let text: string
+
+        // explicit text overrides all other options
+        if (typeof node.text === 'string') {
+          text = node.text
+        } else { // no explicit text
+          const path = node.getAttribute('path')
+          // cross reference points to a file, use the file name
+          if (typeof path === 'string') {
+            text = node.getAttribute('path')
+          } else { // cross reference is an internal reference
+            const refid = node.getAttribute('refid')
+            const refsCatalog = node.getDocument().getRefs()
+
+            // lookup reference by id
+            const refNode = refsCatalog[refid]
+
+            // no reference found for id
+            if (typeof refNode === 'undefined') {
+              text = refid
+            } else { // reference found
+              // maybe the refered node has a reftext which should be used
+              if (refNode.hasAttribute('reftext')) {
+                text = refNode.getReftext()
+              } else { // fall back to title
+                text = refNode.getTitle()
+              }
+            }
+          }
+        }
+        return `<a${attrs.join('')}>${text}</a>`
+      }
     }
     if (nodeName === 'image') {
       const nodeAttributes = node.getAttributes()
--- a/src/test/asciidoctorWebViewConverter.test.ts
+++ b/src/test/asciidoctorWebViewConverter.test.ts
@@ -146,6 +146,97 @@ link:help.adoc[]
 <p><a href="full.adoc" class="bare action button" data-href="full.adoc">full.adoc</a></p>
 </div>`,
     },
+    // xref
+    {
+      title: 'Should resolve "xref:" macro to document',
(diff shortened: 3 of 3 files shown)


preview-src/index.ts
```diff
@@ -82,7 +82,7 @@ const onUpdateView = (() => {
 const updateImageSizes = throttle(() => {
   const imageInfo: { id: string, height: number, width: number }[] = []
   const images = document.getElementsByTagName('img')
-  if (images) {
```
ggrossetie, line 85: is it related to the fix?

src/asciidoctorWebViewConverter.ts
```diff
@@ -202,7 +202,7 @@ export class AsciidoctorWebViewConverter {
       </body>
       </html>`
     }
-    if (nodeName === 'inline_anchor' && node.type === 'link') {
+    if (nodeName === 'inline_anchor' && (node.type === 'link' || node.type === 'xref')) {
```
ggrossetie, line 205: We should implement this logic: https://github.com/asciidoctor/asciidoctor/blob/2983738317abf6bc125c75eede7890c36a10b86c/lib/asciidoctor/converter/html5.rb#L1131-L1154

preview-src/index.ts
```diff
@@ -82,7 +82,7 @@ const onUpdateView = (() => {
...
+  if (images && images.length > 0) {
```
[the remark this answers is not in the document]
ViToni: Indirectly. When looking into the event listener, that clicking on the `xref:` (without any other changes) triggered a `cacheImageSizes` even though there are no images at all in the document I tested with.
It could alos be moved to a dedicated PR.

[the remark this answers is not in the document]
ggrossetie: Indeed, good catch

src/asciidoctorWebViewConverter.ts
ViToni: The Ruby code seems quite complex as it resolves the given `refid`s for non-file `xref:`s. Haven't found code yet which would do something similar. Any hints?

ggrossetie: Asciidoctor.js is transpiled from Ruby to JavaScript. So we should have access to all the Asciidoctor methods.

I can help you implement the non-trivial parts.

ViToni: Currently I have something like this:

```TypeScript
      if (node.type === 'xref') {
        const attrs = []
        if (node.hasAttribute('role')) { attrs.push(` class="${node.getRole()}"`) }

        let text: string
        if (typeof (node.text) === 'string') {
          text = node.text
        }
        const path = node.getAttribute('path')
        if (typeof path === 'string') {
          if (typeof (text) === 'undefined') {
            text = node.getAttribute('path')
          }
        } else {
          // const refid = node.getAttribute('refid')

          // PSEUDO CODE:
          // const refNode = getNodeId(refid)
          //
          // if (refNode.hasAttribute('xrefstyle')) {
          //   const xrefstyle = refNode.getAttribute('xrefstyle')
          //   text = computeText(refnode, xrefstyle)
          // } else {
          //   text = refnode.getTitle()
          // }
        }
        attrs.push(` href="${node.target}"`)
        attrs.push(` data-href="${node.target}"`)
        return `<a${attrs.join()}>${text}</a>`
      }
```

Regular `xref:` to files works quite fine. What I don't know how to solve is the `xref:` to sections in the `PSEUDO CODE` section.
I don't know if there is something like a `computeText` method. I saw that using e.g. `reftext` on sections doesn't change anything in `:toc:`. Don't know if its not handled on purpose or there might be a gap.

ggrossetie: You should be able to get the catalog `node.document.catalog[:refs]`

```js
node.getDocument().getCatalog()['refs']
```

ggrossetie: The Opal/Asciidoctor.js syntax for this part `AbstractNode === ref` is probably a bit complex but you can inspect the `ref`, it should contain a class name.

ViToni:
@ggrossetie I'd say it looks good:

grafik

wdyt?

src/asciidoctorWebViewConverter.ts
ViToni: @ggrossetie I'd say it looks good:

grafik

wdyt?

ViToni: I'd like to add some tests `src/test/asciidoctorWebViewConverter.test.ts` but am not sure how execute them.

ViToni: One can run tests by
```shell
npm run build && node ./dist/src/test/runTest.js
```

ViToni: @ggrossetie If you could spare some time, could you take a look at the tests?
I added test to `src/test/asciidoctorWebViewConverter.test.ts` but the ones referring to files are failing because expected / actual are not matching.
In the actual output all `.adoc` suffixes are replaced by `.html` which is really not expected.

ggrossetie: Thanks for taking the time to fix this issue. I will try to find some time tomorrow to push this fix to the finish line.

ggrossetie: Since we are generating an HTML page, Asciidoctor will use `outfilesuffix` / `relfilesuffix`: 

> Since the default suffix for inter-document cross references in the html5 backend is .html, the resulting link created in these environments may end up pointing to non-existent HTML files. In this case, you need to change the inter-document cross references to refer to other AsciiDoc source files instead.

https://docs.asciidoctor.org/asciidoc/latest/macros/inter-document-xref/#navigating-between-source-files

ViToni: So should I change the test or the backend. As the change is for the preview, I'd assume the backend?

ggrossetie: The preview should use the `.adoc` suffix since we will navigate between AsciiDoc files.

I don't recall if we are already doing it but we should set `:relfilesuffix: .adoc` in the preview.

You can update the convert function in the test to use `:relfilesuffix: .adoc` as well. Your assertions should be valid.

ViToni: > I don't recall if we are already doing it but we should set :relfilesuffix: .adoc in the preview.

This could be the place where `:relfilesuffix: .adoc` is set for the preview:
https://github.com/asciidoctor/asciidoctor-vscode/blob/acbb89000276bc1541b2c66385009f59645b3d60/src/features/asciidoctorAttributesConfig.ts#L18-L23

ggrossetie: Yes that's the one 👍🏻

ggrossetie:
All good, thanks 👍🏻

PR #866, opened 2024-03-31
Convert README / CHANGELOG to AsciiDoc

### LINKED ISSUE #865: Project README / CHANGELOG is Markdown instead of AsciiDoc
The title says it all. The project's README / CHANGELOG should use AsciiDoc instead of Markdown.

ViToni:
This PR converts README / CHANGELOG to AsciiDoc.

Note:
- `git` is able to keep track of renames but only if there are only few changes involved while renaming.
- For this reason the rename of files and the actual changes are in separate commits.
- As this PR has more than only small changes `git`would treat the respective files as "new" when the PR gets squashed.
- Therefore this PR SHOULD NOT be squashed if the history shall be kept from Markdown to AsciiDoc. 

Resolves #865.


ggrossetie:
We can convert the CHANGELOG to AsciiDoc but since https://marketplace.visualstudio.com only supports Markdown, we need to keep the README in Markdown (or use https://github.com/opendevise/downdoc to convert from AsciiDoc to Markdown before publishing to https://marketplace.visualstudio.com/items?itemName=asciidoctor.asciidoctor-vscode).

My goal is to move the documentation to Antora (https://github.com/asciidoctor/asciidoctor-vscode/tree/master/docs) and keep a minimal README (in Mardown or AsciiDoc).
Migrating the documentation to Antora also allows me to test Antora support in the VS Code extension and improve it at the same time as I migrate the documentation 😄

ggrossetie:
I forgot that we are using the CHANGELOG.md content in the GitHub release page so we will also need to convert it from AsciiDoc to Markdown when we create the release. See: https://github.com/asciidoctor/asciidoctor-vscode/blob/master/tasks/release-notes.js + https://github.com/asciidoctor/asciidoctor-vscode/blob/master/release.sh

ViToni:
Thanks for sharing the background information (I was really wondering why the Markdown has been in place for so long 🙈).
Maybe it'd be better to close the ticket and PR then.

PR #872, opened 2024-04-23
Fix document to document xrefs in preview when the documents are included

### LINKED ISSUE #425: Does asciidoc have issues with xref and include?

birdman7260:
Refer to my comment made on the issue for details about the issue and resolution:
https://github.com/asciidoctor/asciidoctor-vscode/issues/425#issuecomment-2071267886

Resolves: in a preview, when a document xref's another document and then that document `include::`'s the first document

The changes made:

- src/asciidocEngine.ts
  - added the intrinsic attributes that were missing. These are missing because they are only intrinsically assigned when asciidoctor processes a file but we process with the input as a string. Reference the asciidoctor documentation, notice that the modifiable column calls out that these are only assignable via the API in this situation: https://github.com/asciidoctor/asciidoctor-vscode/issues/425#issuecomment-2071267886
  - Now that `docname` is set, the `macros` substitution is properly able to handle the xref macros when they are referencing the same document that is `include::`'ing them
- src/test/asciidoctorWebViewConverter.test.ts
  - refactored the two types of conversions to utilize a new function that creates the converter options. This applies the same intrinsic attributes solution
  - also, this sets the safe mode to `UNSAFE` as it is in the preview code, this way `includes::` are actually resolved
  - Added two tests: 1) test that the included file is able to xref into the parent file, 2) new test to make sure an included file can xref a separate included file

resolves #425

### DIFF
--- a/src/asciidocEngine.ts
+++ b/src/asciidocEngine.ts
@@ -1,4 +1,5 @@
 import * as vscode from 'vscode'
+import path from 'path'
 import { AsciidoctorWebViewConverter } from './asciidoctorWebViewConverter'
 import { Asciidoctor } from '@asciidoctor/core'
 import { ExtensionContentSecurityPolicyArbiter } from './security'
@@ -160,11 +161,26 @@ export class AsciidocEngine {
     const antoraSupport = AntoraSupportManager.getInstance(context.workspaceState)
     const antoraAttributes = await antoraSupport.getAttributes(textDocumentUri)
     const baseDir = AsciidocTextDocument.fromTextDocument(textDocument).getBaseDir()
+    let textDocumentExt = path.extname(textDocumentUri.fsPath)
+    const textDocumentName = path.basename(textDocumentUri.fsPath, textDocumentExt)
+    textDocumentExt = textDocumentExt.startsWith('.') ? textDocumentExt.substring(1) : '';
     const templateDirs = this.getTemplateDirs()
     const options: { [key: string]: any } = {
       attributes: {
         ...attributes,
         ...antoraAttributes,
+        // The following attributes are "intrinsic attributes" but they are not set when the input is a string
+        // like we are doing, in that case it is expected that the attributes are set here for the API:
+        // https://docs.asciidoctor.org/asciidoc/latest/attributes/document-attributes-ref/#intrinsic-attributes
+        ...(baseDir && {
+          'docdir': baseDir,// this can be set since safe mode is 'UNSAFE'
+          'docfile': path.resolve(baseDir, `${textDocumentName}.${textDocumentExt}`) // this can be set since safe mode is 'UNSAFE'
+        }),
+        'docfilesuffix': `${textDocumentExt}`,
+        'docname': textDocumentName,
+        'filetype': asciidoctorWebViewConverter.outfilesuffix.substring(1), // remove the leading '.'
+        // end "intrinsic attributes"
+
         '!data-uri': '', // disable data-uri since Asciidoctor.js is unable to read files from a VS Code workspace.
       },
       backend: 'webview-html5',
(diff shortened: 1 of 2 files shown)


birdman7260:
@ggrossetie Let me know if you want me to make a separate issue to tie this to

src/asciidocEngine.ts
```diff
@@ -160,11 +161,26 @@ export class AsciidocEngine {
     const antoraSupport = AntoraSupportManager.getInstance(context.workspaceState)
     const antoraAttributes = await antoraSupport.getAttributes(textDocumentUri)
     const baseDir = AsciidocTextDocument.fromTextDocument(textDocument).getBaseDir()
+    let textDocumentExt = path.extname(textDocumentUri.fsPath)
+    const textDocumentName = path.basename(textDocumentUri.fsPath, textDocumentExt)
+    textDocumentExt = textDocumentExt.startsWith('.') ? textDocumentExt.substring(1) : '';
```
ggrossetie, line 166: I think we should move this code in https://github.com/asciidoctor/asciidoctor-vscode/blob/master/src/asciidocTextDocument.ts

Since `baseDir` is not available in a browser environment (such as https://github.dev) it would be better to handle this case in one location.

birdman7260: I could move that, but I tried to test things in the browser environment and was running into issues.
Is it a known problem that the include directive does not work in the browser environment? Without includes working my fix doesn't actually apply. I tried to use https://github.dev can created a quick example of two adocs and one including the other and it printed out the `Unresolved directive` text in the preview.
Extensions in general are new to me, but I could not figure out how to get it to work, even when I set the `base_dir` according to https://docs.asciidoctor.org/asciidoctor.js/latest/spec/include-support-matrix/
When the runtime is a browser the browser compiled version of asciidoctor is used and it uses XMLHttpRequest to try to open the includes and i just can't figure out how to get that to work correctly... seems the VSCode extension documentation wants all file accesses to use `vscode.workspace.fs` which makes sense. We could probably set up our own `inludeProcessor` like you've done with Antora but I really don't know anything about Antora and whether or not that works in the browser environment. I think it would still be possible to do, might even be nice to make our own `includeProcessor` that works no matter the situation?

That's a fair bit beyond what i can do at the moment though, I see yours is creating Opal classes and whatnot and I'm just barely hangin on when it comes to Opal.

ggrossetie: > Is it a known problem that the include directive does not work in the browser environment?

In theory, it could work if the File System API abstraction provided by VS code can read file in a browser environment.
I don't recall if `include` are currently supported or not in a browser environment.

My point is that we should not execute code that relies on Node.js `path` or `fs`  module since they won't necessarily be available in a browser environment and might throw an error.

For now, you can use `'browser' in process && (process as any).browser === true` to test if we are running in a browser environment and not call `path` or `fs`  module.

I would create a `ComputeIntrinsicAttributesDocumentAttributes` in `asciidocTextDocument.ts` which can return an empty object (or undefined) when running in a browser environment. Does it make sense?

PR #883, opened 2024-05-27
fix asciidoc preview when antora.yml does not specify version

### LINKED ISSUE #882: Antora preview doesn't work if version is not specified in antora.yml
Antora allows using a refname projection as version from the antora playbook (see [here on the antora docs](https://docs.antora.org/antora/latest/playbook/content-source-version/#refname-projection-as-version)).

Earlier [in that page](https://docs.antora.org/antora/latest/playbook/content-source-version/#version-key), the docs specify that 

> If you want the playbook to be able to control the value of the version key, **don’t set the version key in the component version descriptor** (antora.yml file). [bold added for this quote]

**Issue:**
If the version is excluded from the `antoray.yml` file, than the preview in vscode no longer works.

This is similar to the issue reported in #829, reporting that image preview did not work when `version:true` was configured. That issue was  fixed in #871.

alaindresse:
Resolves #882 

Currently, if our antora.yml does not specify a version (required to add a fallback for the version in the antora playbook), it is impossible to preview an asciidoc file.

This pull request extends the approach from @leonardopavanrocha in #871, namely setting the version to an empty string when it is not specified in antora.yml. 

Note:: I also had to bumb the vscode engine number to test this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
 	"browser": "./dist/browser/extension",
 	"main": "./dist/src/extension",
 	"engines": {
-		"vscode": "^1.75.0",
+		"vscode": "^1.89.0",
 		"node": ">=18"
 	},
 	"categories": [
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
 	"browser": "./dist/browser/extension",
 	"main": "./dist/src/extension",
 	"engines": {
-		"vscode": "^1.89.0",
+		"vscode": "^1.75.0",
 		"node": ">=18"
 	},
 	"categories": [
@@ -617,7 +617,7 @@
 		"@types/lodash.throttle": "~4.1",
 		"@types/mocha": "~9.1",
 		"@types/node": "12.20.55",
-		"@types/vscode": "~1.89.0",
+		"@types/vscode": "~1.75.0",
 		"@typescript-eslint/eslint-plugin": "5.62.0",
 		"@typescript-eslint/parser": "5.62.0",
 		"@vscode/test-electron": "2.3.9",
--- a/src/features/antora/antoraSupport.ts
+++ b/src/features/antora/antoraSupport.ts
@@ -18,29 +18,28 @@ export interface AntoraResourceContext {
   module: string;
 }
 
-function getShortHash (longString : string, length = 10) {
-  const hash = longString.split('').reduce((acc, char) => {
-    let code = char.charCodeAt(0)
-    while (code > 0) {
-      acc = ((acc << 5) - acc) + (code % 64)
-      code = Math.floor(code / 64)
-    }
-    return acc
-  }, 0).toString(36).replace(/[^a-z0-9]/g, '').slice(0, length)
-
-  return hash
-}
-
 export class AntoraConfig {
   public contentSourceRootPath: string
   public contentSourceRootFsPath: string
+
+  private static versionMap = new Map<string, number>()
+
   constructor (public uri: vscode.Uri, public config: { [key: string]: any }) {
     const path = uri.path
     this.contentSourceRootPath = path.slice(0, path.lastIndexOf('/'))
     this.contentSourceRootFsPath = ospath.dirname(uri.fsPath)
-    if (config.version === true || !(config.version)) {
-      config.version = getShortHash(path)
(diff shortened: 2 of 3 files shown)


alaindresse:
Changed my code to set version to a short hash of the path of the antora.yml. I was getting issues with duplicate versions for components for which the source was distributed across several folders in the repo.

Use case for the duplication: monorepo with documentation of individual components that I want to consolidate into a single component with appropriate navigation.

ggrossetie:
According to the documentation the `version` is "mandatory" unless inherited from the content source in the playbook. Is that your case?

https://docs.antora.org/antora/latest/component-name-and-version/#antora-yml-requirements

alaindresse:
@ggrossetie thanks for taking the time to look into this despite your busy schedule!

Yes, that is exactly the case:

- If I add the version in the playbook, then it is no longer mandatory in the `antora.yml`, and should actually be absent for the version in the playbook to work.
- If I accordingly remove the version from the `antora.yml`, before my changes, I have to preview in vscode, even though the antora generation works nicely with no errors or warnings.

package.json
```diff
@@ -21,7 +21,7 @@
 	"browser": "./dist/browser/extension",
 	"main": "./dist/src/extension",
 	"engines": {
-		"vscode": "^1.75.0",
```
ggrossetie, line 24: Unless I'm missing something, I don't think we need to bump vscode engine in this pull request?

ggrossetie:
@alaindresse Left a few comments otherwise 👍🏻

ggrossetie:
One test is failing because we assert that the version is `null` but the value is now a random value:

https://github.com/asciidoctor/asciidoctor-vscode/blob/714a0907dc728db15860709814e65152d8a50528/src/test/antoraSupport.test.ts#L188

I don't think we should generate a random value when the version is explicitly `null` with the `~` symbol.

> The `~` values defines a component version as unversioned

https://docs.antora.org/antora/latest/component-version-key/#key

src/features/antora/antoraSupport.ts
```diff
@@ -18,15 +18,28 @@ export interface AntoraResourceContext {
...
+    let code = char.charCodeAt(0)
+    while (code > 0) {
+      acc = ((acc << 5) - acc) + (code % 64)
+      code = Math.floor(code / 64)
+    }
+    return acc
+  }, 0).toString(36).replace(/[^a-z0-9]/g, '').slice(0, length)
+
+  return hash
+}
```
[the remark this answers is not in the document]
alaindresse: Copied it from my personal library, inspired from several sources.

I'll replace it by a more readable fetch into a path - version map.

package.json
alaindresse: Updated based on the following error when packaging:
 ERROR  @types/vscode ~1.89.0 greater than engines.vscode ^1.75.0. Either upgrade engines.vscode or use an older @types/vscode version.

Changing the version of the types to avoid regressions.

alaindresse:
@ggrossetie thanks for your comments.
I believe they are now addressed.

alaindresse:
I ran the tests, using `npm run test` and saw a message saying that all tests passed, but I also saw:

```
Failed to load message bundle for file /Users/alaindresse/code/asciidoctor-vscode/dist/src/features/antora/antoraSupport
Failed to load message bundle for file /Users/alaindresse/code/asciidoctor-vscode/dist/src/asciidoctorWebViewConverter
Failed to load message bundle for file /Users/alaindresse/code/asciidoctor-vscode/dist/src/security
Failed to load message bundle for file /Users/alaindresse/code/asciidoctor-vscode/dist/src/features/documentLinkProvider
```

And did not see any report on the antoraSupport tests.

What is the proper way to run the tests ?

ggrossetie:
@alaindresse Thanks, could you please run `npm i` locally and commit the `package-lock.json`. Apart from that 👍🏻

ggrossetie:
> I ran the tests, using npm run test and saw a message saying that all tests passed, but I also saw:

Message bundle does not properly load during tests (or at least I don't know how to make them work in tests) but it does not prevent tests from running. So this messages can be safely ignored 😅

ggrossetie:
All good 🎉 
Thanks!

PR #886, opened 2024-06-11
mermaid.js support in preview

### LINKED ISSUE #571: Use mermaid.js for diagrams
## Is your feature request related to a problem? Please describe

Having to contact a server (kroki) and also save diagrams on disk (instead of using only plain text) is annoying. Diagrams can't be stored well in git, and I don't want to send my info to kroki, or bother setting up a local server just to make a diagram.

## Describe the solution you'd like

There is already a mermaid.js extension for vs code that creates diagrams from text in Electron in real time, that doesn't need any server or need to save image files. Please add some syntax to allow this extension to be used by the asciidoctor extension. If I have time I'll look into writing this myself.

## Describe alternatives you've considered

No alternative diagram languages that I know of.  They don't have VS code extensions, or are commercially licensed: https://hackernoon.com/my-top-13-javascript-diagram-libraries-g2a53z6u

Also, Mermaid is the default in Github - https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/.

The only real alternative is to use markdown files in VS Code instead of asciidoc files. But then you lose some good features like numbered headings.

ztuowen:
Having to contact a server (kroki) is annoying and I can't use the preview because of security concerns.

This adds a `mini-extension` into the html previewer that processes mermaid diagrams with `<pre class=mermaid>` that `mermaid.js` accepts.

Saving to html feature doesn't work because the saved HTML doesn't load mermaid.js which I haven't figured out how to do.

This will be disabled if the user enables `kroki` which will allow them to save correctly.

resolves #571


ggrossetie:
It looks good thanks!

PR #906, opened 2024-09-25
[FOR DISCUSSION ONLY] - illustrates possible approach for limiting calls to findFiles

alaindresse:
@ggrossetie this shows my explorations on how to limit the impact of calls to findFiles on responsiveness, for discussion.

I will put a few comments in the PR to highlight the points that may be interesting.


package.json
```diff
@@ -673,5 +673,8 @@
 	},
 	"volta": {
 		"node": "20.14.0"
-	}
+	},
+	"enabledApiProposals": [
+		"findFiles2"
```
alaindresse, line 678: Enables use of proposed new findFiles interface, that automatically excludes files that don't need to be searched.

src/features/antora/antoraConfigs.ts
```diff
@@ -0,0 +1,97 @@
+import vscode, { CancellationTokenSource, FileType } from 'vscode'
+import { fastFindFiles } from './fastFindFiles'
+import yaml from 'js-yaml'
+import ospath from 'path'
+
+export class AntoraConfig {
```
alaindresse, line 6: Just copied from antoraSupport

src/features/antora/antoraConfigs.ts
```diff
@@ -0,0 +1,97 @@
...
+    const version = AntoraConfig.versionMap.get(path)
+    if (version) return `V-${version}`
+
+    const nextVersion = AntoraConfig.versionMap.size + 1
+    AntoraConfig.versionMap.set(path, nextVersion)
+    return `V-${nextVersion}`
+  }
+}
+
+const cache = {
```
alaindresse, line 31: This cache is fed at initialisation, and updated when files named `antora.yml` files are changed

src/features/antora/antoraConfigs.ts
```diff
@@ -0,0 +1,97 @@
...
+
+async function refreshAntoraConfigs (token?: vscode.CancellationToken) {
+  if (!ongoingRefreshPromise) {
+    ongoingRefreshPromise = asyncTriggerRefresh(token).then(() => {
+      ongoingRefreshPromise = undefined
+    })
+  }
+  await ongoingRefreshPromise
+
+  async function asyncTriggerRefresh (token: vscode.CancellationToken): Promise<void> {
```
alaindresse, line 46: Triggers refresh of the cache. Logic extracted from antoraSupport.ts

src/features/antora/antoraConfigs.ts
```diff
@@ -0,0 +1,97 @@
...
+
+const cache = {
+  antoraConfigs: [] as AntoraConfig[],
+  antoraConfigUris: [] as vscode.Uri[],
+}
+
+let ongoingRefreshPromise : Promise<void> | undefined
+
+async function refreshAntoraConfigs (token?: vscode.CancellationToken) {
+  if (!ongoingRefreshPromise) {
```
alaindresse, line 39: Used a flag to prevent launching multiple refreshes in parallel: will return the existing refreshPromise if there is an ongoing cache refresh

src/features/antora/antoraConfigs.ts
```diff
@@ -0,0 +1,97 @@
...
+    }))).filter((c) => c) // filter undefined
+
+    cache.antoraConfigs = antoraConfigs
+    cache.antoraConfigUris = antoraConfigUris
+  }
+}
+
+export async function getAntoraConfigs (token?: vscode.CancellationToken) {
+  if (!cache.antoraConfigs.length) {
+    await refreshAntoraConfigs(token)
```
alaindresse, line 77: Only refresh if no cache

src/features/antora/antoraConfigs.ts
```diff
@@ -0,0 +1,97 @@
...
+  return cache.antoraConfigUris
+}
+
+// initial refresh of antora configs
+refreshAntoraConfigs()
+
+// Update the configs when an antora.yml file changes
+vscode.workspace.onDidChangeTextDocument(async (event) => {
+  if (event.document.uri.path.endsWith('antora.yml')) {
+    await refreshAntoraConfigs()
```
alaindresse, line 95: Refresh caches if there is a change in a file called `antora.yml`
We could be smarter and only refresh the info for that specific file, but just wanted to try this out.

src/features/antora/antoraSupport.ts
```diff
@@ -17,32 +26,6 @@ export interface AntoraResourceContext {
   version: string;
   module: string;
 }
-
-export class AntoraConfig {
```
alaindresse, line 21: Moved to antoraConfig.ts

src/features/antora/antoraSupport.ts
```diff
@@ -231,31 +214,6 @@ export async function antoraConfigFileExists (textDocumentUri: Uri): Promise<boo
   return antoraConfig !== undefined
 }
 
-export async function getAntoraConfigs (): Promise<AntoraConfig[]> {
```
alaindresse, line 234: Logic moved to antoraConfig.ts

src/features/antora/fastFindFiles.ts
```diff
@@ -0,0 +1,50 @@
+import vscode from 'vscode'
+
+let findFiles2Available : boolean | undefined
```
alaindresse, line 3: use proposed API if possible

src/features/antora/fastFindFiles.ts
```diff
@@ -0,0 +1,50 @@
...
+
+const exclude : string[] = [
```
alaindresse, line 5: excludes and includes adjusted for performance even if we use the 'standard' findFiles

src/vscode.proposed.findFiles2.d.ts
```diff
@@ -0,0 +1,92 @@
+/* eslint-disable no-irregular-whitespace */
+/* eslint-disable no-undef */
+/* eslint-disable no-tabs */
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+declare module 'vscode' {
```
alaindresse, line 9: Copied to enable the findFiles2 api in codeInsiders

ggrossetie:
I don't want to dig into the implementation but I'm fine with the following changes:

- Make use of `findFiles2` if available
- Add a cache mechanism on `antora.yml` files

For the later, and if we want to be thorough, we will need to add a listener (https://code.visualstudio.com/api/references/vscode-api#FileSystemProvider) and update the cache when:
- a file is created with the name antora.yml
- a file is renamed to antora.yml
- an antora.yml is moved

src/features/antora/fastFindFiles.ts
ggrossetie: I think it's better to use the `files.exclude` configuration as mentioned in the API documentation: https://code.visualstudio.com/api/references/vscode-api#workspace

``**/.DS_Store` and `**/node_modules` are probably fine but I won't be surprised if someone is putting documentation in a directory named `vendor` or `build` (or even `coverage`)

ggrossetie:
Oh, I just found this in the documentation:

> While you're not able to publish extensions using the proposed API on the Marketplace, you can still share your extension with your peers by packaging and sharing your extension.

I guess that would prevent us from using it since we want to publish on the Marketplace :(

alaindresse:
> Oh, I just found this in the documentation:
> 
> > While you're not able to publish extensions using the proposed API on the Marketplace, you can still share your extension with your peers by packaging and sharing your extension.
> 
> https://code.visualstudio.com/api/advanced-topics/using-proposed-api#sharing-extensions-using-the-proposed-api
> 
> I guess that would prevent us from using it since we want to publish on the Marketplace :(

Indeed, I don't believe we will have to wait for the api to be finalized before we can use it (except for internal releases).

alaindresse:
> Could you please open a dedicated pull request for "Use `findFiles2` (proposed API) if available"?

Given the restrictions on publishing an extension that uses a proposed API, I don't believe this is really useful ☹️

alaindresse:
> I don't want to dig into the implementation but I'm fine with the following changes:
> 
> * Make use of `findFiles2` if available
> * Add a cache mechanism on `antora.yml` files
> 
> For the later, and if we want to be thorough, we will need to add a listener (https://code.visualstudio.com/api/references/vscode-api#FileSystemProvider) and update the cache when:
> 
> * a file is created with the name antora.yml
> * a file is renamed to antora.yml
> * an antora.yml is moved

In this proposal, I do listen to `vscode.workspace.onDidChangeTextDocument` (in `antoraConfig.ts` line 93), which I believe would cover `antora.yml` files in the repo. Could you clarify what the FileSystemProvider would add to that ?

When I get some time, I'll try to create a PR that could actually be merged on the caching.

alaindresse:
Closing this pull request as discussion has taken place with clear next steps

- @alaindresse Creates PR on caching antoray.yml locations
- Wait for new findFiles API to change to more efficient searches

ggrossetie:
@alaindresse Could you please share the directory structure of your project? Maybe we could let the user configure the glob pattern to find antora.yml (default value `**/antora.yml`). If you have a common structure, then something like `projects/**/docs/antora.yml` might be faster.

PR #908, opened 2024-09-26
cache config and add options to configure globs

### LINKED ISSUE #907: Antora preview performance issues
Preview update performance is excessively slow (minutes to update a preview on an M1 Pro mac when the source is changed) in large repos with many symlinks (e.g. for nx monorepos with react-native apps and libs).

Using the proposed APIs for findFiles greatly improves the situation, but until that API is finalized, it can only be used with code insiders and it cannot be used in an extension published to the marketplace.

alaindresse:
Caching `antora.yml` configs and adding options to restrict search for `antora.yml` partially helps (but maybe I don't have the right excludes - still looking into that).

partially resolves #907

### DIFF
--- a/package.json
+++ b/package.json
@@ -556,6 +556,16 @@
 						"type": "boolean",
 						"default": false,
 						"markdownDescription": "%asciidoc.antora.enableAntoraSupport.desc%"
+					},
+					"asciidoc.antora.search.antoraConfig": {
+						"type": "array",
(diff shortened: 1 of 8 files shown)


alaindresse:
@ggrossetie this PR implements the improvements we discussed. Not fully tested, not a satisfying solution, but would still appreciate your comments as I continue to search.

Issue #208, opened 2019-06-18
Feature Request - Paste Image


stobias123:
I love this plugin. It's fantastic. Would love to be able to paste images.

## Is your feature request related to a problem? Please describe

I hate that I have to switch contexts and place images appropriately in finder when I'm writing docs. Being able to paste and create appropriate adoc links would be fantastic

## Describe the solution you'd like

Paste creates a link, and uploads the image to `images/`

Seems its possible like this.
https://marketplace.visualstudio.com/items?itemName=mushan.vscode-paste-image

## Describe alternatives you've considered

https://marketplace.visualstudio.com/items?itemName=mushan.vscode-paste-image

I believe this is also possible in the atom asciidoc extensions.

danyill:
@stobias123 You can do this now (actually for a while), using the Paste Image command so I'll close this issue.

![image](https://user-images.githubusercontent.com/674783/74484888-5c0a1800-4f1e-11ea-8e00-372a34ffa5b3.png)

Feel free to re-open if this doesn't address your concern.

PR #925, opened 2024-12-18
update readme with paste image section

leneffets:
Hi there,

just an update on the readme.md. need to stumble over #208 to find out how pasting an image works ;-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ The extension provides a quick command to paste an image from clipboard.
 * Create a screenshot thats in the clipboard
 * Open the command palette - `ctrl+shift+p` or `F1` (Mac: `cmd+shift+p`)
 * Select _AsciiDoc: Paste Image_ (or use the shortcut `ctrl+alt+v`, Mac: `option+cmd+v`)
-* The png file with current timestamp is generated in the same folder
+* The image file with current timestamp is generated in the same folder
 * in the working document an image macro is inserted
 
 ### Snippets
--- a/README.md
+++ b/README.md
@@ -86,6 +86,16 @@ The extension provides a quick command to export your AsciiDoc file as DocBook.
 
 **ℹ️ Note:** Only DocBook 5 is supported.
 
+### Paste Image
+
+The extension provides a quick command to paste an image from clipboard.
+
+* Create a screenshot thats in the clipboard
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -86,6 +86,16 @@ The extension provides a quick command to export your AsciiDoc file as DocBook.
 
 **ℹ️ Note:** Only DocBook 5 is supported.
 
+### Paste Image
+
+The extension provides a quick command to paste an image from clipboard.
+
+* Create a screenshot thats in the clipboard
```
ggrossetie, line 93: Might be too specific? "Copy an image in your clipboard" ?

README.md
```diff
@@ -86,6 +86,16 @@ The extension provides a quick command to export your AsciiDoc file as DocBook.
...
+* Open the command palette - `ctrl+shift+p` or `F1` (Mac: `cmd+shift+p`)
+* Select _AsciiDoc: Paste Image_ (or use the shortcut `ctrl+alt+v`, Mac: `option+cmd+v`)
+* The png file with current timestamp is generated in the same folder
```
ggrossetie, line 96:
```suggestion
* The image file with current timestamp is generated in the same folder
```

README.md
```diff
@@ -86,6 +86,16 @@ The extension provides a quick command to export your AsciiDoc file as DocBook.
...
+* in the working document an image macro is inserted
```
ggrossetie, line 97: The last 2 bullets are not actions that the user must take, so I will probably convert them to a paragraph to describe what the user should see. What do you think?

README.md
leneffets: Yes, great idea! This is the more generic approach. My usecase actually was: take a screenshot and put in in a adoc document. 
But put an image to clipboard is more versatile, should also work e.g. for remote sessions on other hosts.

README.md
ggrossetie: Hmmm good point, you're right we should probably be consistent. Since we are using passive sentence it should be:

An image file, named with a timestamp, is created, and an image macro is added to the document

Anyway, it really is a nitpick. Thanks for updating the documentation, I really appreciate it!

ggrossetie (approved):

PR #986, opened 2026-02-02
fix: use the correct scope for markup.inline.raw

### LINKED ISSUE #342: Syntax highlighting broken for inline code and admonition blocks
## Description

I'm a long-time user of the original **joaompinto.asciidoctor-vscode** extension. With version 2.8.1 of this new extension, syntax highlighting does not properly display for inline code style, and the highlighting for admonition blocks is also different than past behavior. I traced the issue backward through to the old extension and found that the highlighting was fine until **joaompinto.asciidoctor-vscode** version 2.7.10. In that version, the syntax highlighting was updated, and this regression was introduced.

Previously, inline code with single or double backticks would render a different color than normal text, similar to bold/italic text. This is almost certainly expected behavior.

For admonition blocks, the delimiters & content were rendered in a different color than normal text. Now, there is no coloring at all for inline code, and while the title of an admonition block does render nicely, the block itself and delimiters are the same color as regular text. This makes it difficult to see block while scanning a document quickly, but in this case it is less clear what behavior is really expected.

Interestingly, this affects every built-in color theme **except** "Quiet Light", which can properly render the code snippets. Given that, perhaps this is also an issue with VS Code's color theme?

For what it's worth, inline code coloring is a big deal to me, and the admonition blocks are less of an issue.

## System Information

Windows 10 Pro x64
VS Code 1.46.1 (just updated to the latest; also affected me on 1.43.1 before I updated)
Color theme: Monokai
Extension version 2.8.1

## To Reproduce

Create an Asciidoc document with code or admonitions and save it so that syntax highlighting takes effect. Use any of the default color themes except "Quiet Light".

## Screenshots & Files

This is a sample document highlighted with **joaompinto.asciidoctor-vscode** version 2.7.9. Notice the inline code & the admonition block highlighting.

Here is the same sample document highlighted with **asciidoctor.asciidoctor-vscode** version 2.8.1 (which looks the same as **joaompinto.asciidoctor-vscode** version 2.7.10). Notice that the inline code highlighting is gone and the admonition block is no longer rendered differently.

For completeness, here is a copy of the same document rendered with the "Quiet Light" theme. Notice that the inline code is colored correctly.

image

ggrossetie:
resolves #342

### DIFF
--- a/syntaxes/asciidoc.tmLanguage.json
+++ b/syntaxes/asciidoc.tmLanguage.json
@@ -1092,7 +1092,7 @@
               "name": "markup.meta.attribute-list.asciidoc"
             },
             "2": {
-              "name": "markup.raw.monospace.asciidoc"
+              "name": "markup.inline.raw.monospace.asciidoc"
             },
             "3": {
               "name": "punctuation.definition.asciidoc"
@@ -1110,7 +1110,7 @@
               "name": "markup.meta.attribute-list.asciidoc"
             },
             "2": {
-              "name": "markup.raw.monospace.asciidoc"
+              "name": "markup.inline.raw.monospace.asciidoc"
             },
             "3": {
               "name": "punctuation.definition.asciidoc"


abel1502:
When I applied this PR and built the extension from source, it fixed the inline code blocks highlighting, but not the standalone blocks, such as:

```adoc
[NOTE]
====
qqqqqq
====

[source]
----
qqqqqq
----
```

These remain the same color as normal text, except the part in square brackets.

EDIT: Oh, nevermind, code is apparently highlighted with respect to the language if I specify one.

PR #995, opened 2026-03-27
fix(vscode): explicit F5 build tasks and CONTRIBUTING notes

AMDphreak:
## Summary

This PR is intentionally **small**: it only fixes **Run Extension / F5** debugging and documents **why** the old `preLaunchTask` names failed for some setups.

- **`.vscode/tasks.json`** — Define real tasks `dev` and `pretest` that run `npm run dev` and `npm run pretest`, and wire **`launch.json`** `preLaunchTask` to those labels (instead of relying on implicit `npm: dev` / `npm: pretest` tasks from VS Code’s npm Scripts integration).
- **`.vscode/launch.json`** — Same wiring; add **Run Extension (only this extension)** (`--disable-extensions`) for a quieter Extension Development Host when debugging.
- **`CONTRIBUTING.adoc`** — Explain the above in plain language first, then more detail. **npm** remains the default (`npm install`, `npm t`, web test command unchanged). A short **NOTE** mentions **pnpm** for forks that use it—**additional**, not a replacement.

No build-script refactors, no merge from other topic branches, and no change to default npm-oriented contributor flow.

## Plain language

When you press F5, VS Code runs a build task first. The repo used to reference task names like `npm: dev`. Those are **not** script names in `package.json`; they are labels VS Code **may** auto-create. If they are missing, you get `could not find the task 'npm: dev'`. Explicit tasks in `tasks.json` fix that.

## Reviewer detail

The `npm: <script>` pattern came from a **2021** `launch.json` update; checked-in `tasks.json` used a different label (`build`), so the mismatch was easy to miss until implicit tasks disappear (e.g. some package-manager setups).

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -16,7 +16,23 @@
       "outFiles": [
         "${workspaceFolder}/dist/**/*.js"
       ],
-      "preLaunchTask": "npm: dev",
+      "preLaunchTask": "dev",
+      "skipFiles": ["<node_internals>/**"],
+      "showAsyncStacks": true
+    },
+    {
+      "name": "Run Extension (only this extension)",
+      "type": "extensionHost",
+      "request": "launch",
+      "runtimeExecutable": "${execPath}",
+      "args": [
+        "--disable-extensions",
+        "--extensionDevelopmentPath=${workspaceFolder}",
+      ],
+      "outFiles": [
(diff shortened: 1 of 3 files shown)


ggrossetie:
I'm not sure I understand what you're trying to fix. The syntax seems perfectly valid. Please open an issue with the details before creating a PR.

Issue #970, opened 2025-09-08
Add information how to enable syntax highlighting


eirnym:
There is an "obvious" and "well known" information how to enable syntax highlighting in Asciidoc plugin settings using `highlighjs` provided.

For a newcomer when it said "it supports syntax highlighting" , person expects "out of the box" or at least a very easy steps to follow, like a checkbox toggle in settings.

There's few options how to do that:

* In Asciidoc document itself.
* In workspace settings.
* In global user settings.

In Asciidoc document a person need to add:

```asciidoc
:source-highlighter: highlightjs
```

In In workspace or user settings add (or extend with) the similar attirbute:

```json
"asciidoc.preview.asciidoctorAttributes": {
    "source-highlighter": "highlightjs"
  },
```

I'd love to see [this](https://github.com/asciidoctor/asciidoctor-vscode/issues/686) happen as an *option*, but I'd like to keep `highlightjs` for compatibility with other forms of syntax rendering and/or add more syntax highlight engines.

eirnym:
@ggrossetie mat I ask for a new release with this feature?

ggrossetie:
This change will ship as part of the next release. It's a major version with a lot of changes under the hood, so we want to take the time to make sure we're not introducing too many regressions before publishing it. It will be available as soon as it's ready 😉

eirnym:
Could you please share any release issue/milestone to track?

eirnym:
@ggrossetie also I don't see how 2 my issues you've closed recently in #1059 are "major" in either case

Issue #969, opened 2025-09-08
Syntax highlighting is missing a lot of languages from highlightjs


eirnym:
Please include all languages supported by HighlightJS.

*Please provide details about:*

* I try to include `zsh` or other language which is not included in vscode plugin for some reason.
* *I observe:* no syntax highlighing for code block
* *I expect:* highlighting is shown

```asciidoc
[source,zsh]
-------
function fn () {
   result=$(cmd "$@" 2>/dev/null | head -n 1 )
   if [[ $? != 0 ]]; then
     return $?
   fi

   echo "$result"
   return 0
}
-------
```

*How Asciidoc VSCode plugin presented:*

<img width="602" height="251" alt="Image" src="https://github.com/user-attachments/assets/636062bd-d112-4119-a4a6-ca4649cb50f8" />

*How Asciidoc browser plugin and similarly Asciidoc CLI command (highlight js is used) presented the same piece:* 

<img width="414" height="278" alt="Image" src="https://github.com/user-attachments/assets/7d1e485b-27ed-457a-8fa1-a6753970f849" />

Asciidoc VSCode plugin version:   `3.4.4`

*Look among the logs there for errors.*

no plugin logs found

PR #1058, opened 2026-06-24
feat: highlight source code out of the box in the preview (#969, #970)

ggrossetie:
Enable the bundled Highlight.js syntax highlighter by default so source blocks are colored without any configuration.

- Default `source-highlighter` to `highlight.js` (soft `@` default, so a document's own `:source-highlighter:` and the `asciidoc.preview.asciidoctorAttributes` setting still win) in the preview and in HTML-based exports (HTML, wkhtmltopdf). DocBook and the asciidoctor-pdf engine are unaffected (#970).
- Auto-load the Highlight.js grammar for the languages actually used in the document: `highlight.min.js` only registers the ~37 "common" languages, so the preview now detects the source-block languages and loads the matching (already bundled) grammar, making any of the ~190 supported languages work with no `:highlightjs-languages:`. HTML exports still rely on that attribute (#969).
- Document both behaviors and the opt-out/override in the preview page.


ggrossetie:
resolves #970
resolves #969

PR #1180, opened 2026-07-28
fix: load workspace .asciidoctor/lib extensions under native ESM

### LINKED ISSUE #1179: Workspace extension unable to load
I've been testing the pre-release 4.0 version of this plugin with my own extension (here https://github.com/OpenPowerShift/asciidoctor-ldl/releases, version 0.2.1, [asciidoctor-ldl-standalone.js](https://github.com/OpenPowerShift/asciidoctor-ldl/releases/download/v0.2.1/asciidoctor-ldl-standalone.js))

Here is admittedly an AI bug report from Claude but I think it is correct or at least that the workaround resolves the issue.

# Workspace Asciidoctor.js extensions fail to load in 4.0.0 

## Summary

With **Enable Asciidoctor.js extensions registration** (`asciidoc.extensions.registerWorkspaceExtensions`) turned on and a trusted workspace, **every** `.asciidoctor/lib/**/*.js` extension fails to load. The preview shows:

```
c:\…\.asciidoctor\lib\<file>.js: TypeError: Cannot convert undefined or null to object
```

The error is misleading: it names the workspace extension file, but the throw happens **inside the extension's own loader, before the workspace file is ever `require`d**. Any well-formed extension file reproduces it, so the feature is effectively broken in 4.0.0.

## Environment

- asciidoctor.asciidoctor-vscode **4.0.0**
- VS Code (extension host running the extension as native ESM, Node ≥ 20 / Electron)
- `@asciidoctor/core` 4.0.2 (bundled dependency)
- OS: Windows 11 (not OS-specific — it's a module-format issue)

## Root cause

The extension is bundled as native ESM (`dist/src/extension.mjs`). esbuild inserts its standard `__require` shim at the top:

```js
var __require = /* @__PURE__ */ ((x) => typeof require !== "undefined" ? require : typeof Proxy !== "undefined" ? new Proxy(x, {
  get: (a, b) => (typeof require !== "undefined" ? require : a)[b]
}) : x)(function(x) {
  if (typeof require !== "undefined") return require.apply(this, arguments);
  throw Error('Dynamic require of "' + x + '" is not supported');
});
```

In a native ESM module there is no `require` binding, so `typeof require === "undefined"`. The shim therefore falls through to the **`Proxy` fallback**, whose target is the throwing function. Consequently:

- `__require.cache` → the Proxy `get` trap returns `fn["cache"]` → **`undefined`**.
- `__require(path)` → calls the target function → **throws `Dynamic require of "…" is not supported`**.

The workspace-extension loader (`AsciidoctorExtensions.registerExtensionsInWorkspace`) then runs:

```js
for (const extfile of extfiles) {
  const extPath = extfile.fsPath;
  try {
    delete __require.cache[extPath];   // delete (undefined)[extPath]
    const extjs = __require(extPath);  // (never reached)
    extjs.register(registry);
  } catch (e) {
    vscode.window.showErrorMessage(extPath + ": " + e.toString());
  }
}
```

`delete __require.cache[extPath]` evaluates `delete undefined[extPath]`. Resolving the member reference performs `ToObject(undefined)`, which throws:

```
TypeError: Cannot convert undefined or null to object
```

Because `extPath` is interpolated into the caught message, it *looks* like the workspace file failed to load, but the file was never touched.

## Reproduction

1. Any workspace with a valid `.asciidoctor/lib/ldl.js` (e.g. an extension exporting `register(registry)`).
2. Settings: `"asciidoc.extensions.registerWorkspaceExtensions": true`; trust the workspace.
3. Open any `.adoc` preview.
4. Observe: `…\.asciidoctor\lib\ldl.js: TypeError: Cannot convert undefined or null to object`.

Minimal isolation of just the shim, run as native ESM (no `require` in scope — matching the extension host):

```js
var __require = /* @__PURE__ */ ((x) => typeof require !== "undefined" ? require : typeof Proxy !== "undefined" ? new Proxy(x, {
  get: (a, b) => (typeof require !== "undefined" ? require : a)[b]
}) : x)(function (x) {
  if (typeof require !== "undefined") return require.apply(this, arguments);
  throw Error('Dynamic require of "' + x + '" is not supported');
});

console.log(typeof require);      // "undefined"
console.log(__require.cache);     // undefined
delete __require.cache["x"];      // TypeError: Cannot convert undefined or null to object
```

The workspace extension file itself is fine — loading and registering it against `@asciidoctor/core` 4.0.2 (both the `build/node/index.cjs` and ESM `src/index.js` builds) works and renders correctly. The failure is entirely in the loader's use of `__require`.

## Impact

- `asciidoc.extensions.registerWorkspaceExtensions` is non-functional in 4.0.0: no `.asciidoctor/lib/**/*.js` extension can be loaded.
- The error message points at the user's extension file, sending debugging effort in the wrong direction.

## Suggested fix

Give the ESM bundle a real `require` so the esbuild shim resolves to it. Two options:

1. **esbuild banner** (preferred) for the ESM build:

   ```js
   // esbuild config: format 'esm'
   banner: {
     js: "import { createRequire as __cr } from 'module'; const require = __cr(import.meta.url);"
   }
   ```

   With a real `require` in scope, `__require` is the native `require` (functional `.cache`, functional dynamic require).

2. **Load workspace extensions without `require.cache`.** In an ESM context, `require.cache` invalidation isn't available anyway; use `createRequire(import.meta.url)` explicitly in `registerExtensionsInWorkspace`, and drop the `delete …cache…` line (or guard it):

   ```js
   import { createRequire } from 'module';
   const requireExt = createRequire(import.meta.url);
   // …
   delete requireExt.cache?.[extPath];
   const extjs = requireExt(extPath);
   ```

   > Note: re-`require` after cache-delete is how the loader currently picks up edits to a workspace extension between previews. Under ESM that live-reload path needs revisiting regardless.

## Workaround (until fixed)

- Prepend to `dist/src/extension.mjs`:

  ```js
  import { createRequire as __cr } from "module";
  globalThis.require = __cr(import.meta.url);
  ```

  (Verified: restores the loader; `.asciidoctor/lib` extensions load and render. Lost on extension update.)

ggrossetie:
Fixes #1179.

Since 4.0.0 the extension is bundled as native ESM, which has no ambient `require`. `registerExtensionsInWorkspace` still called the bare `require`/`require.cache` to load a trusted workspace's `.asciidoctor/lib/**/*.js` extensions — esbuild's fallback shim for that (a `Proxy` over a throwing stub, since there's no real `require` in scope) threw `TypeError: Cannot convert undefined or null to object` before the extension file was ever loaded, and the catch block then misleadingly attributed the error to the extension file path.

The loader now uses `createRequire(import.meta.url)` (from `node:module`), extracted into a small, vscode-free `requireFresh()` helper (`src/core/requireExtension.ts`) so the fix is directly unit-testable and so a platform without `node:module` (the web extension host) fails per-extension inside the existing try/catch rather than aborting the whole loop. The existing hot-reload-on-edit behaviour (`require.cache` cleared before each reload) is preserved.

The browser build's empty-module shim (`tasks/build-browser.mjs`) gained `node:module` to the list, since the CJS/browser bundle doesn't have a real one either.

Added `src/test/unit/requireExtension.test.ts`, three regression tests: confirms the module runs with no ambient `require` (reproducing the root cause), that `requireFresh` loads a real extension module, and that clearing the cache still picks up an edited file.


mariatorrentedev:
Thanks for the fix! I hit this same error today on 4.0.0 (macOS, Cursor) with a workspace inline-macro extension in `.asciidoctor/lib/`, and traced it independently to the same root cause — the esbuild ESM `__require` shim having no `.cache`, so the loader throws on `delete __require.cache[extPath]` before the extension file is even loaded. Confirming that downgrading to 3.4.5 works as a workaround in the meantime.

@ggrossetie Is there a release planned soon that will include this fix?

ggrossetie:
Probably this week or next week.

PR #1186, opened 2026-08-10
feat: Render exported Mermaid diagrams in browser

### LINKED ISSUE #1170: Mermaid diagrams are not rendered in exported HTML in version 4
#### Description

In version 4.x, Mermaid diagrams are rendered correctly in the VS Code preview, but they are not rendered when exporting the document as HTML.

#### Steps to reproduce

1. Create an AsciiDoc document containing a Mermaid diagram, for example:

```asciidoc
[mermaid]
....
flowchart TD
    A[Start] --> B[End]
....
```

2. Open the AsciiDoc preview and confirm that the diagram is rendered.
3. Run **AsciiDoc: Export Document as HTML**.
4. Open the exported HTML file in a web browser.

#### Actual behavior

The Mermaid diagram is not displayed in the exported HTML.

#### Expected behavior

The exported HTML should render Mermaid diagrams in the same way as the VS Code preview.

#### Possible cause

It appears that the preview and HTML export use different rendering paths.

The preview can render Mermaid diagrams correctly, while the exported HTML does not seem to include everything required to render Mermaid diagrams outside the VS Code preview environment.

#### Possible solutions

Some possible approaches could be:

- Bundle Mermaid.js and the required initialization code into the exported HTML.
- Render Mermaid diagrams to SVG during HTML export and embed the generated SVG.
- Alternatively, delegate Mermaid rendering to Kroki, similar to the approach used in previous versions.

Any of these approaches would make the exported HTML behave consistently with the preview.

#### Environment

- Asciidoctor VS Code extension: 4.x
- Export format: HTML
- Diagram type: Mermaid

YoshihideShirai:
Fixes #1170

## Summary

- keep Mermaid blocks as browser-rendered `<pre class="mermaid">` content in HTML exports
- add Mermaid CDN loading and initialization only when exported HTML contains a Mermaid block
- wrap Mermaid preview output in an `imageblock`-compatible structure so block titles render like Kroki diagrams
- preserve Mermaid block ids and roles on the wrapper while keeping `.mermaid` only on the render target
- model Mermaid diagrams as Asciidoctor `image` blocks internally, then restore them to `<pre class="mermaid">` in preview/export output
- allow tree processors such as `asciidoctor-numbered-captions` to rewrite Mermaid figure captions the same way they do for Kroki image blocks
- cover script injection, no-op, duplicate, class-boundary, title, caption, id, role, and tree-processor caption rewrite cases with unit tests

## Why

Server-side Mermaid SVG rendering can diverge from the preview because jsdom does not provide real browser text layout. Rendering in the browser keeps exported diagrams closer to what users see in the preview and follows the CDN-based approach discussed in #1170.

Mermaid preview blocks were emitted as a bare passthrough `<pre class="mermaid">`, so Asciidoctor block titles were not rendered. Kroki creates image blocks and assigns figure captions, so Mermaid now mirrors that shape for compatible preview and HTML output around titled diagrams.

The passthrough shape also meant downstream Asciidoctor.js tree processors could not reliably adjust Mermaid captions. By keeping Mermaid as an internal image block until conversion time, numbered-caption extensions can see and update the figure caption before the preview/export converter restores the browser-rendered Mermaid source.

## Validation

- `npx @biomejs/biome check src/features/asciidoctor/asciidocEngine.ts src/features/asciidoctor/mermaidExport.ts src/test/unit/mermaidExport.test.ts`
- `npm run typecheck`
- `npm run build-test`
- `npm run test:unit`
- `npm run build-test -- --pretty false`
- `node --test build/src/test/unit/mermaid.test.js build/src/test/unit/mermaidExport.test.js`
- `npm run test:unit -- --test-name-pattern=mermaid`
- `npm run lint`
- `npm run build`
- manual conversion with `asciidoctor-numbered-captions` confirmed a titled Mermaid block renders as `Figure 1-1. My Diagram`


ggrossetie:
Superseded https://github.com/asciidoctor/asciidoctor-vscode/pull/1194

